PR #1, opened 2019-03-26
Fix tests

gsquire:
I saw your post on Lobsters and checked out the code to learn a bit more about Nix. I am on a Mac so I am not sure how to install Nix easily but I believe this fixes the tests.

Perhaps you can make Travis run `cargo test` too?


kolloch:
Awesome, @gsquire!

Thanks a lot! Must have broken it shortly before submitting it.

Unfortunately, I have not worked on building the tests with nix as well.Still need to work on that.

:fireworks:

PR #16, opened 2019-08-23
Change default outputs from Cargo.toml to Cargo.nix

tilpner:
Overwriting Cargo.toml was probably on accident


kolloch:
Yes! Damn, never release in a hurry.

PR #21, opened 2019-08-30
Fix/tidy README

TMVector:
* Seems like you have to access the derivation via the `build` attribute now rather than the `binary` attribute
* Fix one typo ("runnin" -> "running")
* Remove trailing spaces (none were double-spaces for line breaks)


kolloch:
Thank you very much, @TMVector !

PR #29, opened 2019-09-23
Support passing the `ref` attribute to `fetchGit`

### LINKED ISSUE #27: Git depenency only work for default (master) branch
When using a git source (e.g. from a fork of an upstream repository because your changes haven't been merged yet) builds done with crate2nix fail since they can't find the git dependency.

This might not really be a bug in `crate2nix` but rather in `builtins.fetchGit`. As it turned out I can fetch the revision from my fork as soon as the commit is in the default (master) branch.

I did expect `crate2nix` to prefetch the repository and the respective commit while generating the `Cargo.nix` file.

andir:
When fetching a git repository it sometimes is requires to pass the
explicit git branch got `builtins.fetchGit` as it otherwise can't find
the right commit.

Passing the `ref` argument to the function helps in those situations.

Currently only the `branch` attribute of the git URLs is supported. Git
tags should usually also work but I have no reference how those look in
the cargo generated URLs.

Fixes #27

---
This change is [Reviewable](https://reviewable.io/reviews/kolloch/crate2nix/29)

### DIFF
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -151,6 +151,7 @@ pub enum ResolvedSource {
         #[serde(with = "url_serde")]
         url: Url,
         rev: String,
+        r#ref: Option<String>
     },
     LocalDirectory {
         path: PathBuf,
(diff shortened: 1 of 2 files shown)


kolloch:
Thanks for making crate2nix better!

Could you add a test please? E.g. a sample project? If the test glue isn't obvious, I can add that later.

andir:
I'll dig into that later today :)
I am very happy that I can make the code better. Me previous experiences with rust to nix converters wasn't that positive.

PR #30, opened 2019-09-23
Handle special `../` case for path

### LINKED ISSUE #28: Relative path to source not a valid nix path
When generating the `Cargo.nix` into a child directory (e.g. `nix/`) then the relative path to the source code is not a valid nix path.

Example:
```
crate2nix generate -o nix/Cargo.nix
```
Generated expresssion:
```nix
{
# …
        edition = "2018";
        src = (builtins.filterSource sourceFilter ..); # .. should be ../.
# …
}
```
The path should not be `..` but rather `../.`.

andir:
Nix doesn't parse `../` as a valid path it requires a trailing dot in
that case.

fixes #28


kolloch:
Looks great, a test for this would be awesome.

Issue #31, opened 2019-10-04
crate2nix ignores crate-type


asfishman:
The `Cargo.nix` generated by `crate2nix` does not respect the `Cargo.toml` field specifying the crate build type. For example, if the `Cargo.toml` contains the following information:

    [lib]
    crate-type = ["rlib", "staticlib"]

There will be no difference in the `Cargo.nix` output that reflects our desire to build a statically linkable library. Furthermore, the rustc command executed by `buildRustCrate` will not include the desired flags `--crate-type lib --crate-type staticlib`.

The rust machinery in `nixpkgs` already supports this functionality. It just looks like crate2nix does not. I have identified what needs to be added to `Cargo.nix` in order to change this. A crate set just needs to have the attribute `type` which is a list containing one or more of the following strings: 

    bin  lib  rlib  dylib  cdylib  staticlib  proc-macro

If there is no `crate-type` in the `Cargo.toml`, then the `type` attribute of the crate object in `Cargo.nix` should not exist. I manually added this `type` attribute with the appropriate values, and my desired output was achieved.

I have no familiarity with how crate2nix works, so I do not think I will be able to make this feature addition. Hopefully this issue provides sufficient information to make this modification to crate2nix.

asfishman:
I was able to write a quick fix for this. Appreciate feedback:

https://github.com/kolloch/crate2nix/pull/32

andir:
I ran into this as well.
In case someone else needs a workaround for this issue: Simply add the `types` attribute to your crate overrides and it should work (tm).

kolloch:
This was hopefully fixed for lib crate types in #67.

There was already special casing for proc-macros.

PR #32, opened 2019-10-04
Added support for more library types using the prexisting nixpkgs machinery

asfishman:
Looks like I was able to implement a fix for this. I am not sure this is an idiomatic approach as I am relatively new tor rust. I appreciate any feedback.

### DIFF
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -44,6 +44,9 @@ pub struct CrateDerivation {
     pub proc_macro: bool,
     // This derivation builds the root crate or a workspace member.
     pub is_root_or_workspace_member: bool,
+    // The types of targets that should be built. These types are from the following set:
+    // [bin, lib, rlib, dylib, cdylib, staticlib, proc-macro]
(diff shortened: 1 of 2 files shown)


src/resolve.rs
```diff
@@ -95,6 +98,28 @@ impl CrateDerivation {
             .chain(metadata.workspace_members.iter())
             .any(|pkg_id| *pkg_id == package.id);
 
+        let crate_type = package
+            .targets
+            .iter()
+            .fold(Vec::<String>::new(), |types_vec, target| {
```
kolloch: you probably want a `filter/collect` combination instead.

asfishman: I will look into filter and collect.

For more info take a look at my associated issue: https://github.com/kolloch/crate2nix/issues/31

kolloch:
Closing for now. Feel free to follow up.

asfishman:
I don't have the bandwidth to work on this. Feel free to close.

PR #35, opened 2019-10-09
Support building crate2nix on hydra

andir:
This PR adds support to build `crate2nix` on a hydra instance by providing a `nixpkgs` input.
I run my own binary cache that is populated by an hydra instance. By adding a jobset I no longer have to build `crate2nix` on all my machines and in all my shell environments.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -1,0 +1,8 @@
+{ nixpkgs ? null, ... }@_args:
+with builtins;
+let
+  names = filter (n: _args.${n} != null) (attrNames _args);
+  args = listToAttrs (map (name: { inherit name; value = _args.${name}; }) names);in
+{
+  crate2nix = import ./default.nix args;
+}
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,5 @@
-{ pkgs? import ./nixpkgs.nix { config = {}; },
+{ nixpkgs? ./nixpkgs.nix,
+  pkgs? import nixpkgs { config = {}; },
   lib? pkgs.lib,
   cargo? pkgs.cargo,
   nix? pkgs.nix,


release.nix
```diff
@@ -0,0 +1,8 @@
+{ nixpkgs ? null, ... }@_args:
+with builtins;
+let
+  names = filter (n: _args.${n} != null) (attrNames _args);
+  args = listToAttrs (map (name: { inherit name; value = _args.${name}; }) names);in
```
kolloch: This is basically taking all arguments that aren't null, correct? I you added a comment or named the variable `nonNullArgs` it would be a bit clearer to me.

You could also use `filterAttrs` https://github.com/NixOS/nixpkgs/blob/master/lib/attrsets.nix#L124.

Can you explain why that is important for hydra integration? It would be awesome if you could capture that in a comment so that the context is preserved in the future (there is no direct hint in the code why it is needed and what is needed directly to keep hydra working.

What about using https://github.com/NixOS/nixpkgs/blob/master/lib/attrsets.nix#L124 ?

release.nix
```diff
@@ -0,0 +1,8 @@
+{ nixpkgs ? null, ... }@_args:
```
kolloch: What about importing/calling it from ./tests.nix so that we at least have a smoke test from there? (./tests.nix is also "built" when cargo test is run) Kind of imitating what hydra does. That way, I will not inadvertedly break it since I don't use hydra (yet?).

kolloch:
Hey @andir, do you have thoughts on this or should I close it for now?

Do my comments/questions make sense?

kolloch:
Feel free to reopen when ready.

Issue #15, opened 2019-07-23
How to add overrides to `defaultCrateOverrides`


Profpatsch:
The README mentions

> If your library with native dependencies is not yet supported, you can create an overlay to add the needed configuration to the `defaultCrateOverrides`

But it’s not clear to me how that can be done. For example, I can’t find the string `defaultCrateOverrides` in the generated nix file.

kolloch:
Note that this is part of the "standard" `buildRustCrate` support that comes with NixOS. I think it would be very valuable to document this well!

Some rudimentary doc is here: https://git.mayflower.de/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.md#compiling-rust-crates-using-nix-instead-of-cargo

I appreciate pull requests for the `README.md` for crate2nix or the nixos doc itself.

danieldk:
Could you give a hint of how to do this? Normally `buildRustCrate` gives a derivation that is overridable. `override` expects `crateOverrides` as one of the arguments, so that something like the following works:

```
with import <nixpkgs> {};
((import ./hello.nix).hello {}).override {
  crateOverrides = defaultCrateOverrides // {
    hello = attrs: { buildInputs = [ openssl ]; };
  };
}
```

However, the file generated by crate2nix uses `makeOverrideable` again on the derivation (in `buildRustCrateWithFeatures`), for providing the feature overriding functionality. However, this seems to 'override the override', hiding the `buildRustCrate` variant?

Something like the above fails with:

```
error: anonymous function at /home/daniel/.config/nixpkgs/nix-nlp/pkgs/finalfrontier/finalfrontier.nix:2182:8 called with unexpected argument 'crateOverrides', at /home/daniel/git/nixpkgs/lib/customisation.nix:69:12
```

danieldk:
Just to confirm, if I apply this diff to the generated Nix file, I can use `crateOverrides`:

```diff
-  buildRustCrateWithFeatures = {packageId, features}:
-    lib.makeOverridable
-      ({features}: buildRustCrateWithFeaturesImpl {inherit packageId features;})
-      { inherit features; };
+  buildRustCrateWithFeatures = buildRustCrateWithFeaturesImpl;
```

kolloch:
You could also override the defaultCrateOverrides in pkgs with an overlay. Thanks for the detective work.

I'll add crateOverrides to the overridable parameters since it would be an incompatibility otherwise.

PR #38, opened 2019-10-19
Permit overriding buildRustCrate arguments

danieldk:
`buildRustCrateWithFeatures` makes its implementation overridable.
However in doing so, it hides the `override` function of `buildRustCrate`.

With this change attributes other than `features` are passed through to
`buildRustCrate`'s override, so that its arguments are overridable too.

This addresses #15.

---

@kolloch  I made this a draft PR to ask if this is something you would accept. If so, I'll add a test for this as well.

### DIFF
--- a/templates/nix/crate2nix/default.nix
+++ b/templates/nix/crate2nix/default.nix
@@ -68,7 +68,9 @@ rec {
   /* A restricted overridable version of  buildRustCrateWithFeaturesImpl. */
   buildRustCrateWithFeatures = {packageId, features}:
     lib.makeOverridable
-      ({features}: buildRustCrateWithFeaturesImpl {inherit packageId features;})
+      (args@{features, ...}:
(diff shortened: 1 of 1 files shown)


kolloch:
In principle, that looks good to me. Could you provide tests in `tests.nix`, i.e. an example that makes use of that? Thanks.

kolloch:
Oh, and explaining a meaningful example in README.md/CHANGELOG.md would be greatly appreciated (but not required).

danieldk:
Superseded by https://github.com/kolloch/crate2nix/commit/184d3ddb158e47af2014a94b95676036464bf9b7 .

Great to have this functionality now!

kolloch:
I actually didn't think of this when making my change.

I kind of like keeping an interface somewhat restricted so that I know what to test and support. But the generality of your CL has obvious advantages as well.

Thank you.

Issue #18, opened 2019-08-26
Optional crate unavailable


tilpner:
[Testcase here](https://github.com/tilpner/crate2nix-testcases/tree/master/flate-test)

Builds with cargo, not with crate2nix

kolloch:
Thank you very much for reporting this!

Usually build errors are due to differences in the `rustc` command line between `crate2nix` and `cargo`. The things that look like hashes don't matter but feature selection etc. does. If you can isolate that difference, it would help me a lot.

tilpner:
**cargo**
```
rustc --crate-name flate2 /home/tilpner/.cargo/registry/src/github.com-1ecc6299db9ec823/flate2-1.0.11/src/lib.rs --color always --crate-type lib --emit=dep-info,metadata,link -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="miniz-sys"' --cfg 'feature="miniz_oxide"' --cfg 'feature="rust_backend"' -C metadata=f617164663f537ba -C extra-filename=-f617164663f537ba --out-dir /home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps -L dependency=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps --extern crc32fast=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps/libcrc32fast-9cbc2ae3f9c5c1f7.rlib --extern libc=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps/liblibc-2a24069ae9cc47fd.rlib --extern miniz_sys=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps/libminiz_sys-12111c0588ad7829.rlib --extern miniz_oxide=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/deps/libminiz_oxide-7a103db29777ac3d.rlib --cap-lints allow -L native=/home/tilpner/dev/crate2nix-testcases/flate-test/target/debug/build/miniz-sys-68053c9fa6362326/out
```

**crate2nix**
```
[...]

PR #40, opened 2019-10-27
Multi dependency

cchalmers:
This is proposed fix for #18. This makes dependencies as lists instead of attribute sets to allow for multiple dependencies of the same name (with different targets). Admittedly this is not a pretty solution and the generated `Cargo.nix` files have quite a few more lines.

Another option could have been to keep dependencies as attribute sets but make the value a list if there are multiple versions like this:

```
"miniz_oxide" = [
  {
    packageId = "miniz_oxide 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)";
    target = ((target."arch" == "x86_64"));
  }
  {
    packageId = "miniz_oxide 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)";
    optional = true;
  }
];
```

which I've written the nix to make work but I grimaced when I went to edit the template. Maybe it's not so bad or there's another way?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   use of the "renamed crates" feature and therefore could not be build by 
   `buildRustCrate`/`crate2nix` anymore.
 
+* Issue #18 - Optional crate unavailable
+  Allows building packages that have multiple versions of the same dependency (with different
+  targets). In particular the flate2 package now builds.
+
 # 0.5.0 - 0.5.1
 
 Don't use ´Cargo.toml´ but ´Cargo.nix´ as default output! Thank you, @tilpner!
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -71,12 +71,13 @@ rec {
         authors = [
           "Andrew Gallant <jamslam@gmail.com>"
         ];
-        dependencies = {
-          "memchr" = {
+        dependencies = [
+          {
+            name = "memchr";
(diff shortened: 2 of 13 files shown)


jD91mZM2:
This is beautiful. I tried to take a shot at this today myself but gave up pretty quickly, then I found this.

Another potential solution, which is probably worse, would be to change `optional` from the simple boolean that it is now, to rather be able to specify a list of targets on which it is or isn't optional. That way one can merge something like

```toml
miniz_oxide = { version = "0.3.2", optional = true}

[target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
miniz_oxide = "0.3.2"
```

into

```nix
"miniz_oxide" = {
  packageId = "miniz_oxide 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)";
  optional = !((target."arch" == "wasm32") && (!(target."os" == "emscripten")));
};
```

This, however, seems to be complicated as the cfgs parsing is actually invoked from inside the tera templates as far as I can see.

kolloch:
I am so sorry at not getting back to you on this!  This has been lingering for ages and another contributor had a go. I have to wrap my head around this problem first and
I had a bunch of things in my queue and I decided to get ahead with a crate2nix release first. I cannot make promises when I get to this but it would be surprising if it were before next weekend :(

kolloch:
Thanks a lot, this was a great change.

PR #41, opened 2019-11-03
Handle overlapping target.cfg sections better

kristoff3r:
Some packages (e.g. glutin) has the same package in multiple dependency
sections, for example (snippet with other dependencies cut out): 
```
[target.'cfg(target_os = "windows")'.dependencies]
glutin_egl_sys = { version = "0.1.3", path = "../glutin_egl_sys" }

[target.'cfg(any(target_os = "linux", target_os = "freebsd", target_os = "dragonfly", target_os = "netbsd", target_os = "openbsd"))'.dependencies]
glutin_egl_sys = { version = "0.1.3", path = "../glutin_egl_sys" }
```

I messed around a bit and got it to work by extending `target` to `targets` in `ResolvedDependency`, and then combining the conditions with `CfgExpr::Any`. While I think that is strictly better than the current behaviour (which just ignores the second entry), it doesn't handle the cases where they have different features or renames, but that will require some design work.

I tested it locally on an Amethyst project, and it seems to work. If you're interested in merging some variant of this code I can add comments/tests.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -59,23 +59,39 @@ fn cfg_to_nix_expr_filter(
     _args: &HashMap<String, tera::Value>,
 ) -> tera::Result<tera::Value> {
     match value {
-        tera::Value::String(key) => {
-            if key.starts_with("cfg(") && key.ends_with(')') {
-                let cfg = &key[4..key.len() - 1];
+        tera::Value::Array(keys) => {
+            let mut expressions = Vec::new();
+            for key in keys {
+                if let tera::Value::String(key) = key {
+                    if key.starts_with("cfg(") && key.ends_with(')') {
+                        let cfg = &key[4..key.len() - 1];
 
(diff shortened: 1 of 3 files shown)


kristoff3r:
Actually, it looks like we can just merge features as well, because that is what cargo is doing: https://github.com/rust-lang/cargo/issues/4866

kristoff3r:
Oh, I didn't see that #40 does basically the same thing. I should check the other PR's before I start coding :D

kolloch:
I am so sorry at not getting back to you on this! 

I had a bunch of things in my queue and I decided to get ahead with a crate2nix release first. I cannot make promises when I get to this but it would be surprising if it were before next weekend :(

kristoff3r:
After looking at and testing #40, I can confirm that it is a better version of this one. I hope it gets merged soon.

Also, no worries about the response time. And thanks for creating this useful project!

PR #48, opened 2019-11-15
Fix double-escape of nix strings for build dependencies

andir:
Currently `crate2nix` generates quoted `target` expressions in the `buildDependencies` attribute set, e.g.:
```nix
{ # …
        buildDependencies = {
          "cc" = {
            packageId = "cc 1.0.47 (registry+https://github.com/rust-lang/crates.io-index)";
            target = "(target.\"os\" == \"dragonfly\")"; # <---- the quotes should not be here
          };
        };
# …
}
```

With this PR I am adding a test case and a fix that marks those strings as safe as they were already properly escaped previously.


crate2nix/templates/build.nix.tera
```diff
@@ -168,7 +168,7 @@ rec {
             usesDefaultFeatures = false;
             {%- endif -%}
             {%- if dependency.target %}
-            target = {{dependency.target | safe | cfg_to_nix_expr}};
+            target = {{dependency.target | safe | cfg_to_nix_expr | safe }};
```
kolloch: I don't think that the first safe is still needed, this should be sufficient:

```
            target = {{dependency.target | cfg_to_nix_expr | safe }};
```

Am I missing something?

kolloch:
Merging and correcting afterwards...

PR #54, opened 2019-11-25
Add fuchsia as an (unsupported) target

jamii:
Some crates (eg getrandom) generate a `target."fuschia"` expression, which currently throws an error.


kolloch:
Thanks! A small test would be appreciated.

kolloch:
(just to prevent regressions)

PR #55, opened 2019-11-25
Add support for fetching submodules, using nix-prefetch-git

### LINKED ISSUE #34: Git submodules missing / Switch to pkgs.fetchgit
When using a git source for a crate the git submodules are not being fetched. This is a limitation of `builtins.fetchGit`. By using `pkgs.fetchgit` one can fetch those but is then required to provide the output hash of the source archive.

In the long run it is probably best to just do something like `nix-prefetch-git` for all of the git crates and always use `pkgs.fetchgit` (or even `pkgs.fetchFromGitHub` where possible).

This would also aid generating expressions that can be built in pure mode. For example on hydra instances.

cpcloud:
Closes #34.


crate2nix/src/prefetch.rs
```diff
@@ -93,22 +114,51 @@ pub fn nix_prefetch_from_crate_io(
...
+            url.as_str(),
+            "--fetch-submodules",
+            "--rev",
+            rev,
+        ];
+
+        // TODO: --branch-name isn't documented in nix-prefetch-git --help
+        // TODO: Consider the case when ref *isn't* a branch. You have to pass
+        // that to `--rev` instead. This seems like limitation of nix-prefetch-git.
+        if let Some(r#ref) = r#ref {
```
kolloch: Could this produce a wrong hash then somehow?

Could we fall back and to not returning the hash and caching that prefetch doesn't work?

regenerate_cargo_nix.sh
```diff
@@ -6,18 +6,20 @@ echo "================ Regenerating ./Cargo.nix =================="
...
      { echo "Bootstrap regeneration of ./Cargo.nix failed." >&2 ; exit 1; }
 
-nix-shell --run 'crate2nix "generate" "-n" "./nixpkgs.nix" \
-  "-f" "./crate2nix/Cargo.toml" "-o" "./crate2nix/Cargo.nix"'  ||\
+nix run -c crate2nix generate -n ./nixpkgs.nix \
+  -f ./crate2nix/Cargo.toml -o ./crate2nix/Cargo.nix || \
      { echo "Regeneration of ./Cargo.nix failed." >&2 ; exit 1; }
 
-crate2nix=$(nix-build)/bin/crate2nix
+nix build
```
kolloch: why is that better?

regenerate_cargo_nix.sh
```diff
@@ -6,18 +6,20 @@ echo "================ Regenerating ./Cargo.nix =================="
...
 
-nix eval --json -f ./tests.nix buildTestConfigs |\
- jq -r .[].pregeneratedBuild |\
+crate2nix="$(pwd)"/result/bin/crate2nix
+
+nix eval --json -f ./tests.nix buildTestConfigs | \
+ nix run nixpkgs.jq -c jq -r .[].pregeneratedBuild | \
```
kolloch: nice :)

crate2nix/src/prefetch.rs
cpcloud: I don't think it can return an invalid hash, but it may cause the prefetch command to fail if something other than a branch is used for the `r#ref`. Currently that can't be the case IIUC, but I suspect we want to allow tags in the future.

PR #57, opened 2019-11-29
Fix README: stable -> unstable

bbigras:
I guess this from a copy/paste. Maybe you prefer 'master' or something else instead of 'unstable'.


kolloch:
Thank you :)

PR #60, opened 2019-12-11
Add (basic) support for running tests

andir:
This is a first step in starting to support running tests of crates in the current workspace.

As part of this I had to start refactoring `target` as that was being passed around as global state since it is required to override `target.test` in order to get features for test builds.

While doing that I decided to refactor some function calls from "normal" function arguments to attribute sets as arguments. The pattern of what was being passed on and required became a bit repetitive so it seemed like a good idea.

The idea is to pass `doTest = true` (e.g. using `.override`) to the build to additionally calculate the test features and dependencies. The build of that crate will then be passed to `buildRustCrate` (from nixpkgs) where it will be used to build & execute the tests in that crate. My current idea is to have those tests "build" as a dedicated derivation that will be required for the actual crate to succeed.

While this doesn't fully close that chapter yet it is a first step and some basic implementation to discuss how we want this to evolve. I'll  draft a PR to nixpkgs next that adds support for testing crates.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -68,6 +68,44 @@ rec {
       baseName == "tests.nix"
     );
 
+  /* Returns a crate which depends on successful test execution of crate given as the second argument */
+  crateWithTest = crate: testCrate:
+    let
+      # override the `crate` so that it will build and execute tests instead of
+      # building the actual lib and bin targets We just have to pass `--test`
+      # to rustc and it will do the right thing.  We execute the tests and copy
+      # their log and the test executables to $out for later inspection.
+      test = (testCrate.overrideAttrs (old: {
+        name = "${old.name}-test";
+        postBuild = ''
+          mkdir $TMP/tests
+          find target/lib target/bin -type f -executable -exec cp {} $TMP/tests/ \;
+          for file in $TMP/tests/*; do
+            echo "Executing test $file"
+            "$file" 2>&1 >> $TMP/tests.log || exit 1
+          done
+        '';
+        # remove any other outputs, this also purposely breaks using this as a
+        # rust library in case someone gets confused and passes this into
+        # `dependencies` of `buildRustCrate`.
+        outputs = [ "out" ];
+        installPhase = ''
+          mkdir -p $out
+          cp -rv $TMP/tests $out
+          mv $TMP/tests.log $out
+        '';
+      })).override {
+        extraRustcOpts = [ "--test" ];
+      };
+    in crate.overrideAttrs (old: {
+      checkPhase = ''
+        test -e ${test}
+      '';
+      passthru = (old.passthru or {}) // {
+        inherit test;
+      };
+    });
+
   /* A restricted overridable version of  buildRustCrateWithFeaturesImpl. */
   buildRustCrateWithFeatures =
     { packageId
@@ -78,10 +116,18 @@ rec {
     }:
     lib.makeOverridable
       ({features, crateOverrides, doTest}:
-        let builtRustCrates = builtRustCratesWithFeatures {
-          inherit packageId features crateOverrides buildRustCrateFunc doTest;
-        };
-        in builtRustCrates.${packageId})
+        let
+          builtRustCrates = builtRustCratesWithFeatures {
+            inherit packageId features crateOverrides buildRustCrateFunc;
+            doTest = false;
+          };
+          builtTestRustCrates = builtRustCratesWithFeatures {
+            inherit packageId features crateOverrides buildRustCrateFunc;
+            doTest = true;
+          };
+          drv = builtRustCrates.${packageId};
+          testDrv = builtTestRustCrates.${packageId};
+        in if doTest then crateWithTest drv testDrv else drv)
       { inherit features crateOverrides doTest; };
 
   /* Returns a buildRustCrate derivation for the given packageId and features. */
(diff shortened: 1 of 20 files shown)


sample_projects/bin_with_git_branch_dep/Cargo.nix
```diff
@@ -361,14 +392,14 @@ rec {
     in cacheWithAll;
 
   /* Returns the enabled dependencies given the enabled features. */
-  filterEnabledDependencies = dependencies: features:
+  filterEnabledDependencies = {dependencies, features, target}:
     assert (builtins.isList dependencies);
     assert (builtins.isList features);
```
kolloch: Also add an assert for target here?

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -70,65 +69,88 @@ rec {
...
+    , crateConfigs ? crates
+    , crateOverrides
+    , buildRustCrateFunc
+    , doTest
+    , target ? defaultTarget
+    } @ args:
     assert (builtins.isAttrs crateConfigs);
     assert (builtins.isString packageId);
     assert (builtins.isList features);
+    assert (builtins.isAttrs target);
```
kolloch: What about doTest?

I also think that deserves some mention in the function comment. Is it a bool? What does it enable?

Also, is it a customary name for running tests in nix that other build generators use? Otherwise, I'd prefer `runTests`.

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -70,65 +69,88 @@ rec {
...
+
+    let mergedFeatures = mergePackageFeatures ( args // { inherit target; });
+        mergedTestFeatures = mergePackageFeatures ( args // { target = target // { test = true; }; });
+
+        buildByPackageId = packageId: buildByPackageIdImpl packageId false;
+        buildTestByPackageId = packageId: buildByPackageIdImpl packageId true;
```
kolloch: This is still unused, correct?

tests.nix
```diff
@@ -98,13 +98,23 @@ let crate2nix = pkgs.callPackage ./default.nix {};
...
+            cargoToml = "Cargo.toml";
+            expectedOutput = "Hello, cfg-test!";
+            pregeneratedBuild = "sample_projects/cfg-test/Cargo.nix";
+          }
+
+         {
+            name = "sample_project_cfg_test-with-tests";
+            src = ./sample_projects/cfg-test;
+            cargoToml = "Cargo.toml";
+            expectedOutput = "Hello, cfg-test!";
```
kolloch: Can you change cfg-test in such a way that you actually notice the difference? E.g. use annotations to print something else?

andir: Yeah, I could probably use some kind of 

```
#[cfg(test)]
println!("foo");
```

crate2nix/templates/nix/crate2nix/default.nix
andir: In nixpkgs (and stdenv) there are a few "phases" that match what we are going to do here:

- checkPhase
- installCheckPhase

and the boolean flag for that is something like `doCheck`.

I have no opinions on this option. I think going with something around `check` is probably best if we aim for nixpkgs compatibility.

crate2nix/templates/nix/crate2nix/default.nix
andir: > This is super cool, thank you.

Thank you for all the work that I didn't have to do :heart: 

> For merging (now), I would, for now, either make `doTest` do something or remove it. (Or did I miss that it already does something? ) Then we can already merge and you'll have less potential merge conflicts down the way.

I would probably postpone merging just a few more days until I've the nixpkgs change drafted. Then I have all the pieces to present and we can probably get a much better initial interface and do not have to worry about backwards compatibility (for now).

> I guess that if "doTest" is true, we should optimally create a dervivation that depends both on the release build and the tests -- that way they can build in parallel:
> 
>     * crate derivation, outputs include release binaries and maybe as a different output (never used multiple outputs) the test results
>       
>       * release build derivation with output binaries (what we have currently)
>         
>         * release dependencies ....
>       * run tests derivation that runs the tests, output are the test results
>         
>         * a derivation for each test?
>           
>           * test dependencies ...

My plan was to inject a version of `rootCrate` that has `target.test=true` enabled into `buildRustCrate`. The contract here would be that additionally passed derivation resembles a configuration of the crate that has test support enabled.
`buildRustCrate` could then either override that derivation to fit its needs (extraRustcArgs = [ "--test" ];, modified install phase, added check phase that runs all the tests). On the other hand we could also implement most of that in `crate2nix`. I am currently looking more towards having it in nixpkgs as that allows us to have a common interface among many rust2nix tools.


> Given how buildRustCrate is currently structured (information is passed from build.rs to the rest of the build via env variables), this might be difficult to achieve. For example, I think one would need to capture the results of build.rs so that they are available in other derivations without rerunning build.rs.

I did that as an experiment some time ago and that wasn't too hard. I basically added a few arguments to `build.rs` that allowed running tests instead of producing outputs.

Because that was rather hacky I started reworking the entire `buildRustCrate` expression. The exact interface I am aiming for isn't set yet. I believe the implemention can be enhanced incrementally as long as we have a proper interface.

> 
> Thus, we should probably aim lower and just execute tests in the same derivation that we build the test binaries? Anyway, if there is potential for easy splitting into multiple derivations there, I am all for it.

As it stands every rust crate can have an entirely different set of dependencies during testing. I would try to avoid mixing both of them and thus in favor of having tests run in a dedicated expression.

andir:
Ok, I just pushed a few commits that I produced on my journey to actually get tests to work.



I went from a simple (hack) that wouldn't need any kind of nixpkgs support to a change in Nixpkgs that makes things so much easier. The non-nixpkgs version did work fine until I had crates the did require build scripts generated by `buid.rs`. With the nixpkgs version I was able to just properly declare that I want to build tests and only amend some of the rustc flags - all of those that are actually needed vs all of them.

I intend to squash these commits together so that the intermediate hack isn't visible in the final version. Those changes are just there so people can get an idea how it would look like if done otherwise.

Initially I was hoping that we could just have some test support in `crate2nix` and iterate on the test infrastructure before adding that to `nixpkgs`. The other option would have been to move the entire `buildRustCrate` infrastructure into the generate `default.nix` which would be very bloated but very flexible.

The major downside of the test support based on `nixpkgs`'s `buildRustCrate` is that we are unlikely to be able to support building rust crates in any stable `nixpkgs` release before the 20.03 release has been made. IMO that isn't that big of a deal since you can just use a newer nixpkgs for your crates and do not have to switch your entire system to it.

Since executing tests is still opt-in that might also not be a big deal. We just must communicate that to the users of `crate2nix` that if they want test support they need at least a specific minimum nixpkgs version.

For all of this to work the following nixpkgs branch is required: https://github.com/andir/nixpkgs/commits/buildRustCrate-test
Since the nixpkgs work is derived from the previously mentioned PR I did not open a PR just yet. I'll do that shortly after that change goes in.

kolloch:
Hi @andir,

wouldn't it make sense to first merge the pull request for nixpkgs? Or is that what you want to do anyways?

And are you at the CCC congress? then you can reach me with a DM at https://twitter.com/pkolloch. I can reply with telegram/WhatsApp contact info...

andir:
> wouldn't it make sense to first merge the pull request for nixpkgs? Or is that what you want to do anyways?

That is what I am aiming for. I just wanted to post this here in case you have very different opinions on this.

I just opened the nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/76820

andir:
I rebased this (onto #63) and addressed the feedback I got. Now that https://github.com/NixOS/nixpkgs/pull/76820 is merged we can probably move ahead here?

kolloch:
Hi Andir, thanks a lot for the amazing work.

I merged it but I consider the functionality still experimental in the sense that I want to play a bit with it and maybe think about the nix interface to this functionality more. That might result in some changes but maybe also not.

PR #67, opened 2020-01-21
Support multiple lib crate types

### LINKED ISSUE #31: crate2nix ignores crate-type

andir:
As of now we did only support building rust shared .rlib's or binaries.
A common case is to build shared libraries that will be called from
C/C++/Python/… code. To make those work we need to follow the
`crate-type` specified in the Cargo.toml.

This potentially fixes #31


kolloch:
Hi Andreas, 

thanks a lot for yet another PR! I restricted and simplified the logic a bit. Also I changed it so that `type = [ "lib" ]` will not be rendered, since it is the default? Didn't look into the logic of the crate types too much so I hope my assumptions are correct.

Thank you for also providing a test.

andir:
> thanks a lot for yet another PR! I restricted and simplified the logic a bit. Also I changed it so that `type = [ "lib" ]` will not be rendered, since it is the default? 

Yeah that is the default (at least in `buildRustCrate`). It is probably fine. I just didn't want to have a dependency on the defaults between both projects.

PR #72, opened 2020-01-23
Add parallel downloads and indicatif progress bar

### LINKED ISSUE #68: Prefetch mutliple crates in parallel and/or asynchronously
When running `crate2nix generate` currently the tool appears to be prefetching the dependencies sequentially and struggles to saturate the bandwidth. Parallelizing (or making it asynchronous scatter-gather) would speed up this tool significantly.

cpcloud:
Closes #68.


cpcloud:
There still may be an issue with the denominator value. With the progress bar here I'm seeing maximum numerator values that are less than, equal to, and greater than the denominator.

kolloch:
Nice stuff!

I just started refactoring things this morning.

What do you prefer, integrate your changes on top of mine, once I am done, our shall I just use your ideas? The latter will probably not be a proper rebase since I changed too many things and therefore loose your name...

cpcloud:
I'll wait for your changes, as I just rewrote this PR to use async instead of rayon and it appears to be slightly faster.

cpcloud:
I'm going to split this into a couple of PRs

kolloch:
Alright, my change is in.

cpcloud:
Sweet, will redo my PR then. Thanks for the update!

cpcloud:
@kolloch This is ready for review.

crate2nix/src/lib.rs
```diff
@@ -152,11 +152,6 @@ fn prefetch_and_fill_crates_sha256(
     }) {
         if let Some(hash) = lock_file.get_hash(&package.package_id.repr)? {
             package.source = package.source.with_sha256(hash);
-        } else {
```
kolloch: Why did you remove this? I'd like to include it so that it is clear when this is happening if the Cargo.lock format should change.

cpcloud: It seemed a bit noisy to me. Would you be open to a PR that adds logging with the ability to choose at runtime the level, so that say this message would only show up at the `Debug` level?

cpcloud: Happy to bring this back and open up another PR to add logging.

cpcloud:
@kolloch Is there any particular reason why we need to mutate `CrateDerivation`s instead of constructing new instances? I think constructing new instances would simplify the prefetching code quite a lot.

crate2nix/src/lib.rs
cpcloud: Ok, brought it back.

crate2nix/src/prefetch.rs
```diff
@@ -97,62 +104,87 @@ pub fn prefetch(
...
+                            None
+                        },
+                    )
+                }))
             }
+        },
+    );
+
+    // TODO: Is there a better way to choose this number?
+    let n_concurrent_tasks = num_cpus::get() * 10;
```
kolloch: This does not really depend on the CPUs -- more on network bandwidth and the speed of the server at the other end. Therefore, this seems to be excessive. I don't think that 160 is a reasonable number of parallel prefetches, yet this can happen with this auto detection.

I would rather go with "4" by default or something like that and make it configurable via command line options.

crate2nix/src/prefetch.rs
```diff
@@ -165,13 +198,14 @@ fn get_command_output(cmd: &str, args: &[&str]) -> Result<String, Error> {
         .map_err(|_e| format_err!("output of '{} {}' is not UTF8!", cmd, args.join(" ")))
 }
 
-trait PrefetchableSource: ToString {
+#[async_trait]
```
kolloch: I fear that bringing in async into this project - as is - has no advantage but adds a lot of dependencies.

We are still spawning a process for every fetch...

kolloch:
@cpcloud About mutating CrateDerivation instead of returning copies.

"Historically", I was curious to explore using mutability in a safe way in rust as opposed to using a traditional functional approach.

What I don't like about the current approach is that the `CrateDerivation` struct is quite unwieldly and therefore it is not clear with the types what is actually mutated. Not sure if creating new `CrateDerivation` objects is better here.

I would love to better encapsulate the prefetching so I can also use it without `cargo metadata` but only the lock file. I guess it would end up having an "immutable" interface taking a Vec<ResolvedSource> and returning something like Map<ResolvedSource, String> where the string is a prefetched hash.

Tell me what you think.

cpcloud:
> For me, parallel downloads is not an essential feature but a nice to have.

We have a very large set of dependencies that takes incredibly long to download without parallel downloads. We actually just migrated our codebase from separate independent crates to a single workspace that contains all crates. In the former scheme without parallel downloads crate2nix became untenable to use because it would takes many minutes to prefetch. Even with the workspace, with parallel downloads we have to prefetch around 450 dependencies.

crate2nix/src/prefetch.rs
cpcloud: Yes, but the process is completely IO bound - one place where async shines. I'm actually using the single-threaded scheduler in this implementation precisely because the computation is IO bound and there would be nothing beneficial about scheduling a bunch of threads that simply wait on IO.

I might be missing something though, I'm relatively new to Rust and to async for that matter.

crate2nix/src/prefetch.rs
cpcloud: I agree that using num cpus here doesn't make sense, and configuration is a good idea. However, I think we still want a relatively large number of tasks by default (tens of tasks).

If one download is slow then we can still make progress on others (of course, only up to a point).

160 might not be reasonable, but I think eg 60 would be.

I can measure roughly where the point of diminishing returns is (recognizing that this will be dependent mostly on network factors)

lovesegfault:
As @cpcloud's coworker I will say that prefetching ~700 deps yesterday took forever, it'd be great if it was done in parallel due to that.

I'll add that it makes perfect sense to fetch many (>>ncpu) crates at the same time, but I'm also not sure how to determine that number, maybe look into how cargo does it (since it also does parallel prefetch).

Finally, regarding indicatif, while per-crate bars would be nice I think it's significantly harder to achieve than the "total" progress bar, which by itself is already a huge improvement in UI over the status quo.

cpcloud:
> About mutating CrateDerivation instead of returning copies.

Safe mutability is nice, but no mutability is better IMO as I find that a safely mutable object is no easier to use than one that isn't safely mutable.

Is the safe mutation out of concern for performance, or purely a curiosity related choice?

Prefetch without cargo metadata would be sweet. I'd have to look more at the code to see where the separation should exist.

kolloch:
Thank you for the discussion!

About mutability: I already said about mutability what you are asking for. I don't think that refactoring it to an immutable version but return a new CrateDerivation for everything is making the code better (for me). Performance on that level should not matter for crate2nix if another solution where better.

About async: Yes, by using tokio, you might make crate2nix a miniscule amount more performant  while introducing complicated dependencies. If the network fetches were performed in crate2nix itself (which is probably not really a good idea because it would bloat the code), I'd see your point. But this way, we save some threads and then create a new process for every fetch. The new process cost will be bigger than creating a thread (both relatively small) and typically, the actual execution time of the fetching will dominate everything else.

About a larger default for parallel prefetching: Well, what can I say, I am really happy of introducing a command line parameter for this so that you can use huge values. I am not happy with making the default large. I don't want crate2nix to turn into a load testing tool for underpowered systems and networks.

Which leads me to the most interesting question: How is your project set up? If we assume that the fetch time is dominated by network bandwidth, which I guess is usually true, the parallelism doesn't matter that much. Did you measure? If you fetch stuff from many different slow servers, it might.

Then, when are you prefetching sources? And why so many?

* If you check in `crate-hashes.json`, your checksums should be cached. The only reason not doing this that I can think of are common merge conflicts if there is a lot  churn (we could find solutions for that). Therefore, only new dependencies need to be prefetched. 
* Well, if really a lot of your dependencies change all the time, then there are still a lot of prefetches. Is that the case?
* Then in the master version, crate2nix tries to get the hashes from the `Cargo.lock` file for dependencies from crates.io. I suspect that this does not work for you because you removed the warning that gets printed. I'd love to understand why,

kolloch:
Thanks for the discussion BTW :+1:

cpcloud:
> Yes, by using tokio, you might make crate2nix a miniscule amount more performant while introducing complicated dependencies.

See below for the performance improvement. To my eye it appears to be a significant improvement.

Regarding dependencies, I think you're right about the complexity if `crate2nix` were a library since in that scenario you're forcing your users to take on the new deps as well. However, `crate2nix` is an application and the dependency burden rests almost solely on the maintainers of and contributors to `crate2nix`. It's on us to make sure we understand the dependencies of the code and their implications. Downstream users just want the application to be reliable and fast and typically aren't worried about whether some command line tool they use takes on a new dependency.

Moreover, the `tokio` dependencies I've introduced here are lightweight: the core _single-threaded_ runtime and the `async` `Process` API.

---

After measuring, the network bandwidth doesn't appear to be the issue. Bandwidth is never even close to being saturated.

Here's the progress bar for a version **without** `tokio::process::Command`:
```
[00:03:41] [########################################] 458/458
```

And the progress bar **with** `tokio::process::Command`:
```
[00:00:15] [########################################] 458/458
```
That's a nearly _15x_ performance improvement simply by converting the primary download loop to use async.

---

> Then, when are you prefetching sources? And why so many?

IMO this question misses the point: regardless of why and how many there will be cases like ours where we have a lot of dependencies that need to be prefetched. Users should have a good experience when using `crate2nix` no matter what, especially if the only thing required to support it is a smattering of `async` and a tiny bit of `tokio`.

kolloch:
Nice benchmark! I am not against prefetching in parallel but I think it should be fine with threads if this stays the only use case for async. That would then also use the stable std interface and not add another unstable dependency. I would be very surprised if it made a measurable difference in speed.

In any case, I am interested in your setup. If it is possible for you to elaborate, please humor me. Note that I'd like to accomodate all reasonable use cases but if you don't tell me yours, I cannot.

I also want to understand **why** you have a 15x speed up. Did you try with different degrees of parallelism?

kolloch:
Maybe you can elaborate as well, @lovesegfault?

lovesegfault:
@kolloch We have a number (52) of internal crates, and they are all part of one big workspace, and thus share one lock file. Some of these internal crates have a large number of dependencies (transient or immediate) which adds up to about 701 total dependencies, all of which are prefetched with crate2nix.

Moreover, we do _not_ check in `crate-hashes.json` in our repo in order to avoid collisions (~40 devs touching this codebase / cargo workspace every day).

cpcloud:
@kolloch To clarify, what executor would you like me to use? There's one in tokio, but there's also one in the futures crate, and there's also `async-std` as well. Do you have a preference? I personally don't mind which one we use here so I happily defer to your preference.

cpcloud:
> Did you try with different degrees of parallelism?
Yes, 120 was slightly better than 60 which was _much_ better than using the number of cores. I also tried this without using `tokio::process`, using multiple threads and that was just as slow as master is. I also tried that with differing degrees of parallelism as well and they were all terribly slow.

There doesn't appear to be anything other than tokio that supports asynchronous subprocesses AFAICT. `async-std` is missing that API at the moment (https://docs.rs/async-std/1.4.0/async_std/process/index.html)

I'd really like to land this feature if possible as it would greatly improve the lives of my colleagues.

Is there any other information you need?

lovesegfault:
I just benchmarked this PR vs. `master`:

`master`:
```
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
12.31user 3.26system 2:56.08elapsed 8%CPU (0avgtext+0avgdata 47372maxresident)k
0inputs+2528outputs (0major+1026895minor)pagefaults 0swaps
```

PR:
```
[00:00:11] [########################################] 457/457 (0s)
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
16.35user 4.05system 0:12.88elapsed 158%CPU (0avgtext+0avgdata 49700maxresident)k
0inputs+2528outputs (0major+909631minor)pagefaults 0swaps
```

This is a **14.6x** improvement over master.

lovesegfault:
This is the same test but using my phone as a hotspot, to attempt and constrain the network. This was on an iPhone 11 Pro, on 4G LTE from Verizon, hotspot is over WiFi (as opposed to cabled)
```
❯ nix run nixpkgs.speedtest-cli -c speedtest
[1 copied (0.1 MiB), 0.0 MiB DL]
Retrieving speedtest.net configuration...
Testing from Verizon Wireless (174.194.147.146)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Wiline Networks (San Francisco, CA) [15.59 km]: 40.523 ms
Testing download speed................................................................................
Download: 71.63 Mbit/s
Testing upload speed......................................................................................................
Upload: 12.68 Mbit/s
```

`master`:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
12.28user 3.22system 6:04.49elapsed 4%CPU (0avgtext+0avgdata 47456maxresident)k
0inputs+2776outputs (0major+1027063minor)pagefaults 0swaps
```

PR:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
[00:00:20] [########################################] 457/457 (0s)
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
15.92user 3.85system 0:23.96elapsed 82%CPU (0avgtext+0avgdata 109176maxresident)k
8inputs+21944outputs (1major+920202minor)pagefaults 0swaps
```

Here we have a **15.21x** improvement, with an even more jarring overall time difference.

lovesegfault:
This is a benchmark run on our office using a workstation ([lshw](https://gist.github.com/8e96ec340a1b5158ea20bdc961400ab5)) connected over SFP (10Gbps) to a fiber switch connecting to our ISP.
```
❯ nix run nixpkgs.speedtest-cli -c speedtest
Retrieving speedtest.net configuration...
Testing from Webpass (136.25.38.10)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Wiline Networks (San Francisco, CA) [1.20 km]: 1.651 ms
Testing download speed................................................................................
Download: 937.19 Mbit/s
Testing upload speed......................................................................................................
Upload: 929.90 Mbit/s
```

`master`:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
14.20user 5.16system 3:09.10elapsed 10%CPU (0avgtext+0avgdata 43472maxresident)k
589992inputs+2489outputs (3275major+1024529minor)pagefaults 0swaps
```

PR:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
[00:00:08] [########################################] 457/457 (0s)
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
17.33user 4.79system 0:09.71elapsed 227%CPU (0avgtext+0avgdata 44244maxresident)k
513356inputs+2491outputs (0major+909338minor)pagefaults 0swaps
```

lovesegfault:
Lastly, this is a comparative benchmark run on a RaspberryPi 4 ([lshw](https://gist.github.com/e989425db902bbb971d24e9de10dd5bd)) connected to our office network via WiFi. I hope this illustrates the worst combo of slow network, slow disk, and slow cpu.

```
❯ nix run nixpkgs.speedtest-cli -c speedtest
Retrieving speedtest.net configuration...
Testing from Webpass (136.25.38.10)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Wiline Networks (San Francisco, CA) [1.20 km]: 5.879 ms
Testing download speed................................................................................
Download: 50.63 Mbit/s
Testing upload speed......................................................................................................
Upload: 66.57 Mbit/s
```

`master`:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
56.02user 18.83system 4:04.74elapsed 30%CPU (0avgtext+0avgdata 59344maxresident)k
3768inputs+824176outputs (17major+1027931minor)pagefaults 0swaps
```

PR:
```
❯ rm -f crate-hashes.json && nix run -f nix/nixpkgs.nix update-crate2nix -c time update-crate2nix
...
[00:00:34] [########################################] 457/457 (0s)
Wrote hashes to ./crate-hashes.json.
Generated ./Cargo.nix successfully.
63.47user 20.52system 0:36.81elapsed 228%CPU (0avgtext+0avgdata 47508maxresident)k
0inputs+215448outputs (0major+884885minor)pagefaults 0swaps
```

**6.65x** improvement

kolloch:
Hi @lovesegfault / @cpcloud ,

I see now that it matters a lot for you. I suspect by what @lovesegfault told me that you use `rust nightly` and therefore using the checksums directly from the lock file did not work. This should be fixed. Or are those large numbers of crates not from creates.io?

Anyways, for people having a large number of git dependencies (maybe you), this might still make a huge difference. Is this the case? Are you having a lot of git dependencies?

I am sympathetic with adding parallelism but without adding dependencies like "tokio" to crate2nix if this is their only use. I tried to say that before but I think I was misunderstood. I'd rather keep the number of dependencies of crate2nix smaller. And I don't think that the difference will be noticeable.

So I'd like to just use `std` threads. No progress bar, no tokio, ... I can code that later when get around it, or you do it.

I see that you already introduce a config option. I'd like to lower the default value to `25`. If you need more, you can increase it manually. A good dynamic default might be the nix configuration option `http-connections` as reported by `nix show-config`.  But maybe that's too complicated.

lovesegfault:
@kolloch 

That's correct, we are on nightly which we bump every once in a while. There are also a handful of git dependencies, less than 5 I think, so I think those shouldn't matter too much.

I have to say that I disagree completely with the idea that hand-rolling your own progress indicator is better than using a shared library. I understand with the wish of keeping dependencies down, but I can't comprehend refusing QoL improvements because of dependencies. It's clear from the benchmarks that the cost of the extra handful of dependencies crate2nix would acquire is largely outweighed by the difference in usability.

Any way, it's your prerogative, of course.

kolloch:
Maybe I am missing something obvious here, so it is not a rethoric question: How is a progress bar a big improvement over the current state of affairs?

Like I said, I do see the point of parallelization now. Thanks to this discussion. Even though I think that the fixed Cargo.lock support and the issue where I propose a merge conflict free crate-hashes.json #82 will improve your life much more noticeably.

That's my opinion for now. It can be wrong and it can change but I need to make decisions, right? ;)

Thanks for the discussion.

kolloch:
I am sorry that I made encouraging noises the beginning and then looked at it in detail and thought about it and changed my mind.

I don't think it was wasted effort since I learned a lot about Crate2nix is used and it will get better.

cpcloud:
@kolloch I'm not sure what to think here. I can rewrite the PR to use std::thread, and we can get parallel downloads merged. Are you now saying that you're against parallel downloads?

nagisa:
It is the time for the *first ever* invocation of `crate2nix` that they are concerned here the most. As we don’t check in crate-hashes, it is going to download *all* of the dependencies on a clean check-out *regardless* of the compiler version or other external factors – just to re-generate said crate-hashes file. Having to wait 5+ minutes for this tool to run is just… extremely bad user experience.

The reason downloads are slow stem mostly from the overhead in HTTP. There are multiple roundtrips required even before the data is transferred and each crate is usually in the range of a few dozen kilobytes. This effectively means that most of those 5 minutes is spent waiting on these roundtrips to happen, rather than the actually useful job of downloading the crate data. Just to draw an analogy, regular websites have to deal with exactly this same issue. It was solved by implementing something called [HTTP pipelining](https://en.wikipedia.org/wiki/HTTP_pipelining).

Parallelization as implemented in this PR effectively implements HTTP pipelining. Taking the asynchronous implementation approach effectively ensures that the *runtime* overhead of doing so is minimal (threads being way more expensive than tasks).

While @cpcloud shouldn’t have conflated the progress bar and parallel downloading mechanisms into the same PR, I think outright closing this PR on *both* counts is just disappointing and perhaps an indication of misguided priorities for this project.

nagisa:
FWIW one great alternative implementation would be to use HTTP2 multiplexing, but that would involve pulling in `hyper` which in turn would pull in the whole async runtime regardless.

cpcloud:
> FWIW one great alternative implementation would be to use HTTP2 multiplexing, but that would involve pulling in `hyper` which in turn would pull in the whole async runtime regardless.

Indeed, this is what `cargo` does albeit using `curl` instead of `hyper`.

kolloch:
My position again:

* Parallel downloads are fine, just don't add tons of dependencies to `crate2nix`. Stuff in std should suffice.
* Make number of parallel invocations small by default and configurable.

This will make the case in which you add a lot of new dependencies at once faster. 

After #82, there should be no reason, not to check in your crate-hashes. Do you disagree? Checking them in will give you a dramatically better experience then any optimizations in parallel fetching (which I still accept). As will getting the checksums from Cargo.lock which now works for 1.41.

I spent a lot of time in this discussion. Please take time to try to understand my position as well. If you want to convince me, be kind, try to answer my questions. 

I already told you about my reasoning why threads are in the same order of magnitude of costs. We are still forking a process for every fetch which is more expensive than a thread. This is still right, so please don't repeat that tasks are cheaper. It doesn't matter. 

Honestly, it feels like you are group arguing against me, no matter what. Everything that I know about your situation seems to hint, that your situation will improve way beyond what this PR achieved on its own. A thank you would be appreciated.

nagisa:
Sorry, I noticed that this is spawning `nix-prefetch-*` for each downloads only after writing that comment. Should have retracted it immediately after I did, but for some reason I didn’t get around to it.

I think given the implementation of fetching it does indeed not make much sense to tack on async just to wait on a couple of processes and you’re right in rejecting it. Sorry for ganging up on you and thanks for not giving in ;)

Since we’re spawning processes, I think the ideal approach here is to just be smart about how we `Process::wait` for them. I will try to sketch up a much simplified PR shortly.

PR #78, opened 2020-01-29
Fix test runner

nagisa:
The part that comes after `--` is the `FILTER` and thus impedes the
passing of any other flags properly. More over, doing `"$foo"` makes
shell interpret all the flags as a single argument which is also wrong
for extra flag lists that have more than a single element.


crate2nix/Cargo.nix
```diff
@@ -3597,10 +3597,9 @@ rec {
       in pkgs.runCommand "run-tests-${testCrate.name}" {
         inherit testCrateFlags;
       } ''
-        set -e
+        set -ex
         for file in ${drv}/tests/*; do
-          echo "Executing test $file" | tee -a $out
-          $file -- "$testCrateFlags" 2>&1 | tee -a $out
+          $file --test $testCrateFlags 2>&1 | tee -a $out
```
kolloch: Shouldn't we keep the "--" here?

@andir

andir:
I am a bit confused why you think the tests wouldn't be run otherwise. They clearly are being executed for me. Also the tests of this project should fail if the tests aren't being executed. Is there something wrong in there as well?

nagisa:
Here are some outputs for some combinations of flags (note filtered out)

```
$ /nix/store/h64wp3lxs1xirrkmf08vf6k2az0amlnv-some-test/tests/some-0d0155d7e2 | grep 'test result'
test result: ok. 88 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ /nix/store/h64wp3lxs1xirrkmf08vf6k2az0amlnv-some-test/tests/some-0d0155d7e2 -- --skip definitely_does_not_match_anything | grep 'test result'
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 88 filtered out
$ /nix/store/h64wp3lxs1xirrkmf08vf6k2az0amlnv-some-test/tests/some-0d0155d7e2 --test --skip definitely_does_not_match_anything | grep 'test result'
test result: ok. 88 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ /nix/store/h64wp3lxs1xirrkmf08vf6k2az0amlnv-some-test/tests/some-0d0155d7e2 --test -- "--skip definitely_does_not_match_anything" | grep 'test result'
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 88 filtered out
$ /nix/store/h64wp3lxs1xirrkmf08vf6k2az0amlnv-some-0.1.0-test/tests/some-0d0155d7e2 --test "--skip definitely_does_not_match_anything" | grep 'test result'
error: Unrecognized option: 'skip definitely_does_not_match_anything'
```

nagisa:
That being said, `--` is definitely not the right thing to have there and the additional flags must not be quoted.

andir:
Ok, so the problem here is not that we are adding `--` that doesn't actually matter. Granted, it might be wrong in this case. What matters is that we do not pass `--skip foo` as one argument. We must pass it in unquoted and that is the actual change that matters here.

`--test` is the default action those test binaries perform. We can be explicit about that and pass `--test`. 

Can you update the commit message to reflect what we both figured out now? That would make a lot more sense since strictly speaking tests were being run before that change. The flags (skip, …) were just not (always?) being passed correctly.

```
$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test

running 1 test
test echo_foo_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test --test

running 1 test
test echo_foo_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test -- --skip echo

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test --skip echo

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test -- --skip echo

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

$ /nix/store/hnzc0apz3rsil89xjplpwi11w04rr5ri-rust_cfg-test-0.1.0-test/tests/tests_echo_foo_test --test -- --skip echo

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

```

andir:
We might want to add a regression test for this case as well.

It should be as simple as adding `#[test] fn failing_test() { assert!(false); }` and ignoring that in the respective test for the crate.

kolloch:
Thank you both, you are awesome! Is one of you up to writing a regression test? I would love that very much.

nagisa:
Added the test.

sample_projects/test_flag_passing/test.nix
```diff
@@ -0,0 +1,10 @@
+{ pkgs? import ../../nixpkgs.nix { config = {}; }
+, generatedBuild ? ./Cargo.nix { } }:
+
+let instantiatedBuild = pkgs.callPackage generatedBuild {};
+in instantiatedBuild.rootCrate.build.override {
+    runTests = true;
+    testCrateFlags = [
+      "--skip" "this_must_be_skipped"
```
andir: Shouldn't this be in one argument as otherwise it would also work before this change?

nagisa: Before this change the test would have failed regardless of this being a single (`--skip this_must_be_skipped` interpreted as the pattern in its entirety and all tests are skipped) or multiple arguments (`--skip` is interpreted as the pattern and no functions match, `this_must_be_skipped` is ignored entirely).

nagisa:
```
$/tmp/xxx/target/debug/deps/xxx-04aa82fa44a00818 -- '--skip' 'it_works'

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

$/tmp/xxx/target/debug/deps/xxx-04aa82fa44a00818 -- '--skip it_works'

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out
```

kolloch:
I don't know if you saw that but there is some build failure...

nagisa:
Well it was cachix’s fault so I deleted it. You’re welcome.

kolloch:
Hi nagisa! Thanks for looking into it.

I'll have a look.

kolloch:
Can you remove the commit removing cachix? I want to test if it works now and I cannot update your branch apparently.

kolloch:
Oh, I will merge. Maybe I'll even remove cachix. Too bad that it is unreliable

PR #80, opened 2020-01-31
README.md: add cargo2nix

eupn:


kolloch:
Maybe elaborate a bit more. What is special about it, how does it differ.  I think it has some cross building facilities, right?

Also, would you return the favor and include a link to crate2nix in that project?

Generally, I am also really open to work together where we share the same goals.

kolloch:
I will close this for now to mark it as "reviewed" for me. Reopen when you added details etc.

PR #84, opened 2020-02-01
Sketch out an alternative parallel prefetch algo

nagisa:
This commit sketches out the alternative proposal for prefetching
sources in parallel. The trick used was to split up the `prefetch`
function for the sources into two separate `start_prefetch` and
`finish_prefetch` functions.

With those we can maintain a simple queue to ensure we never run more
than the specified number of jobs. That being said this implementation
has a deficiency in that it has head-of-the-line blocking problem (i.e.
if the earliest spawned pre-fetch blocks for a long time, and later ones
finish earlier, we fail to spawn more work.

I still think its an improvement of status quo, especially given how
little complexity it introduces.

cc @cpcloud @lovesegfault

### DIFF
--- a/crate2nix/src/lib.rs
+++ b/crate2nix/src/lib.rs
@@ -155,7 +155,8 @@ fn prefetch_and_fill_crates_sha256(
         *hash = nix_base32::to_nix_base32(&bytes);
     }
 
-    let prefetched = prefetch::prefetch(config, &from_lock_file, &default_nix.crates)
+    let prefetched = prefetch::Prefetcher::new(config, &from_lock_file, &default_nix.crates)
+        .prefetch()
         .map_err(|e| format_err!("while prefetching crates for calculating sha256: {}", e))?;
 
     for package in default_nix.crates.iter_mut() {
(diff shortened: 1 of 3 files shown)


crate2nix/src/lib.rs
```diff
@@ -224,4 +224,5 @@ pub struct GenerateConfig {
     pub crate_hashes_json: PathBuf,
     pub nixpkgs_path: String,
     pub other_metadata_options: Vec<String>,
+    pub jobs: usize,
```
kolloch: I think that we should make the name to include "prefetch", i.e. "parallel_prefetches" because, for me, jobs is misleeding.

crate2nix/src/main.rs
```diff
@@ -71,6 +71,14 @@ pub enum Opt {
             help = "The path to the crate hash cache file. Uses 'crate-hashes.json' in the same directory as Cargo.toml by default."
         )]
         crate_hashes: Option<PathBuf>,
+
+        #[structopt(
+            short = "j",
```
kolloch: This strongly implies something about the build since it is like the make argument. I would like to make clear that this is the number of parallel prefetches.

nagisa:
@kolloch I went ahead and split up the function into smaller pieces. I fiddled with adding a test or two for this, but realized that adding them would take more time than I have to dedicate for this project right now, so I will not be adding them in this PR.

kolloch:
I merged it on `origin/master` in the `nagisa-parprefetch` branch.

`./run_tests.sh` had some easy to fix compile errors and a bunch of clippy warnings. 

It seems that this has dropped in your priority, If you want to follow up on this, let me know.

Issue #86, opened 2020-02-06
Add `@generated` to header


lovesegfault:
Tools such as Phabricator, look for `@generated` to identify that changes to a file shouldn't be shown in a diff view.

There's precedent in Cargo itself to do this: https://github.com/rust-lang/cargo/issues/6180

PR #87, opened 2020-02-06
templates: rephrase to include `@generated`

lovesegfault:
Closes: #86


kolloch:
Thank you. This is in affect merged but I merged it manually in a way that is not recognized by github.

PR #89, opened 2020-02-07
Run tests from crate source directory

nagisa:
Crates written assuming cargo may also assume that the current working
directory is set to crate root when tests are being run.

This will make strictly more tests pass, although tests which write to
$PWD will still fail to run. This could be rectified by making a view in
`/tmp` or something, but its way beyond me to fix at this point in time.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -83,6 +83,7 @@ rec {
         inherit testCrateFlags;
       } ''
         set -ex
+        cd ${crate.src}
         for file in ${drv}/tests/*; do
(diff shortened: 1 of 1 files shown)


crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -83,6 +83,7 @@ rec {
         inherit testCrateFlags;
       } ''
         set -ex
+        cd ${crate.src}
```
andir: I think that would work for most cases.

There is probably a case where someone relies on stuff that is being generated by the `build.rs` script... In any case this sounds good to go in. We should add a test that makes use of this tho..

kolloch:
Thanks!

If you find a time, I would love to have a test for this.

PR #95, opened 2020-02-10
regenerate_cargo_nix.sh: fix with channels

alyssais:
This workaround is required when using NIX_PATH pointing to a
directory, rather than individual named elements[1].  The fix[2] still
hasn't made it into a release.

[1]: https://github.com/NixOS/nix/issues/1865
[2]: https://github.com/NixOS/nix/commit/1d5cb6ad4839a50a96c27c94f19adcb97b6391af


kolloch:
Thank you, @alyssais! :)

Issue #94, opened 2020-02-10
Verify/Fix `defaultCrateOverrides` argument in the Cargo.nix build file


kolloch:
The README contains an example of how to override the `defaultCrateOverrides` which allows customizing the buildRustCrate parameters of individual crates: https://github.com/kolloch/crate2nix#patching-crate-derivations-with-crateoverride

There is also a build file argument `defaultCrateOverrides` which should set the default for the whole file.

* [ ] Create a test case that it actually works. Probably by adding a sample_project and use in in `./tests.nix`.  You will need a test with a `customBuild`.
* [ ] If it doesn't work, fix it.

PR #97, opened 2020-02-10
Fix crateOverrides

alyssais:
This seems to have been implemented at some time in the past, but got
lost, since the documentation talks about crateOverrides, and the
crateOverrides variable is passed around only for nothing to happen
with it.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -3690,7 +3690,9 @@ rec {
         buildByPackageIdImpl = packageId:
           let
             features = mergedFeatures."${packageId}" or [];
-            crateConfig' = crateConfigs."${packageId}";
+            unoverriddenCrateConfig = crateConfigs."${packageId}";
+            crateOverride = crateOverrides.${unoverriddenCrateConfig.crateName} or (lib.const {});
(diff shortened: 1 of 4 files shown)


kolloch:
Funny, I just created https://github.com/kolloch/crate2nix/issues/94 :)

kolloch:
The intention was to pass it to `buildRustCrate` which accepts a [`crateOverrides`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/default-crate-overrides.nix). Also see https://github.com/kolloch/crate2nix#patching-crate-derivations-with-crateoverride and https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/default-crate-overrides.nix.

I think that is still a good idea. And this time, we should definitely add an integration test for that.

kolloch:
Oh, I am obviously also happy about README.md corrections and/or issues if something is unclear or wrong.

kolloch:
I'll probably have a go at fixing this. Let me know if you'd rather do it! @alyssais

kolloch:
I looked into it and came up with a fix.

b567d8cd2c78c5c37ca34ecda317b87d8c179215
https://github.com/kolloch/crate2nix/issues/94

PR #107, opened 2020-02-18
Support providing `buildinputs` to test execution.

andir:
This allows users to add `buildInputs` to the test execution. This might
be required as during test execution external programs will be called on
generated output or in cases where the rust application is just a
wrapper around some other tool (in some way or another).

### DIFF
--- a/sample_projects/bin_with_git_branch_dep/Cargo.nix
+++ b/sample_projects/bin_with_git_branch_dep/Cargo.nix
@@ -184,10 +184,15 @@ rec {
         || baseName == "tests.nix"
       );
 
-  /* Returns a crate which depends on successful test execution 
-     of crate given as the second argument. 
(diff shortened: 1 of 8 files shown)


sample_projects/bin_with_git_branch_dep/Cargo.nix
```diff
@@ -187,7 +187,7 @@ rec {
   /* Returns a crate which depends on successful test execution 
      of crate given as the second argument. 
   */
-  crateWithTest = crate: testCrate: testCrateFlags:
+  crateWithTest = crate: testCrate: testCrateFlags: testInputs:
```
kolloch: I'd prefer an attrset argument here. And some parameter docs and type asserts.

andir:
I added a few lines about running tests to the README.md and (hopefully) addressed all of other review comments.

PR #117, opened 2020-03-04
Parameterize local source fetcher

pikajude:
The project I'm using `crate2nix` for already has some stuff set up for source filtering using `git ls-files` and so on, so this makes it a lot easier to use that as a drop-in replacement for the bundled source filter.

### DIFF
--- a/crate2nix/templates/build.nix.tera
+++ b/crate2nix/templates/build.nix.tera
@@ -15,6 +15,7 @@
 , rootFeatures ? [ "default" ]
   # If true, throw errors instead of issueing deprecation warnings.
 , strictDeprecation ? false
+, getCleanedSource ? null
 }:
 
 rec {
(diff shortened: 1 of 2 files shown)


crate2nix/templates/build.nix.tera
```diff
@@ -15,6 +15,7 @@
 , rootFeatures ? [ "default" ]
   # If true, throw errors instead of issueing deprecation warnings.
 , strictDeprecation ? false
+, getCleanedSource ? null
```
andir: Please add a comment here explaining the usage of the attribute.

kolloch:
One idea that I also head is to check for the presence of "source.nix" or "nix/source.nix" file and use that instead if present.

But that would require "import from derivation". So just dropping here in case that seems interesting.

crate2nix/templates/build.nix.tera
kolloch: Something in the README.md would also be appreciated!

kolloch:
Hi @pikajude, feel free to follow up on this! I am closing this for now.

PR #123, opened 2020-03-25
crateWithTests: simulate a cargo-like file layout when running tests

symphorien:
This enable seamless execution of integration tests based on cli_test_dir, like nix-du's.

Tested by compiling nix-du, which requires https://github.com/NixOS/nixpkgs/pull/83379.

I tried running the tests but they fail on ssl problems...

```
=============== Regenerating sample_projects/bin_with_git_branch_dep/Cargo.nix ================
Generated sample_projects/bin_with_git_branch_dep/Cargo.nix successfully.
=============== Regenerating sample_projects/sub_dir_crates/Cargo.nix ================
Prefetching    1/1: https://github.com/kolloch/with_sub_crates.git#f8ad2b98ff0eb5fea4962f55e3ced5b0b5afe973
Initialized empty Git repository in /run/user/1000/git-checkout-tmp-pgmKUwwj/with_sub_crates-f8ad2b9/.git/
fatal: unable to access 'https://github.com/kolloch/with_sub_crates.git/': SSL certificate problem: unable to get local issuer certificate
fatal: unable to access 'https://github.com/kolloch/with_sub_crates.git/': SSL certificate problem: unable to get local issuer certificate
Unable to checkout f8ad2b98ff0eb5fea4962f55e3ced5b0b5afe973 from https://github.com/kolloch/with_sub_crates.git.
Error: while prefetching crates for calculating sha256: nix-prefetch-git
=> exited with: 1
Regeneration of sample_projects/sub_dir_crates/Cargo.nix failed.
==================
/home/symphorien/src/crate2nix/regenerate_cargo_nix.sh : FAILED
```


crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -111,9 +111,30 @@ rec {
...
         } ''
           set -ex
-          cd ${crate.src}
+
+          export RUST_BACKTRACE=1
+
+          # recreate a file hierarchy as when running tests with cargo
+
+          # the source for test data
+          cp -r ${crate.src}/* .
```
andir: I this really beneficial over the old `cd ${crate.src}`?

symphorien:
Yes, this way, the source is writable.

symphorien:
(as an example of repos where this might be useful, I think rustfmt tests automatically patch the test cases when there is a breaking change in formatting)

kolloch:
Hi @symphorien, thanks for your PR! 

Since you write "might be useful", did you encounter this in practice or just did it on a gut feeling that it might be useful? My understanding would be that generally rust code should avoid changing the sources directly. If you want to generate sources, you usually do that in `build.rs` and that in turn should only write into `$OUT_DIR`.

Unfortunately, I also don't understand your example:

>(as an example of repos where this might be useful, I think rustfmt tests automatically patch the test cases when there is a breaking change in formatting)

Might be a great example, I am just not getting what you mean.

Usually, the source is not thaaat big, so copying it should be fine but still, if it's just on a hunch, I'd rather not. Also, usually you are mostly interested running tests on crates under your control where you could change the code.

kolloch:
> I tried running the tests but they fail on ssl problems...
> 
> [...]
> fatal: unable to access 'https://github.com/kolloch/with_sub_crates.git/': SSL certificate problem: unable to get local issuer certificate

That sucks! Thanks for trying to run the tests. I think that I reproduced the issue and could fix it by adding the `cacert` dependency to the dev dependencies. Could you try to rerun the tests on master, please? @symphorien

I could reproduce by cleanly checking out the sources again. Apparently, somehow my `direnv`/`lorri` setup worked.

symphorien:
This was just a hunch, but it is actually corroborated by the code:

https://github.com/rust-lang/rustfmt/blob/rust-1.26.0/tests/lib.rs#L208-L209

They run rustfmt on rustfmt's source. When I packaged rustfmt for nixos, tests were failing so these files would actually be modified. If I had been using crate2nix, instead of having the intended informative error message of the test suite, which prints the diff iirc, I would only get weird permission errors unreproducible outside of nix-builds.

symphorien:
I rebased on master and tests pass !

boxofrox:
> This was just a hunch, but it is actually corroborated by the code:

FYI.  That code does not appear to overwrite any files in the source.  Instead, it looks like this test collects errors and warnings generated by `rustfmt` when checking files in its own source tree.

All calls to `format_input` (e.g. [L383 \[1\]](https://github.com/rust-lang/rustfmt/blob/rust-1.26.0/tests/lib.rs#L383)), pass `None` as the output stream/file, so no files appear to be changed.

If there's some other indirect function call that modifies the source tree, I don't see it upon first glance.

\[1]: https://github.com/rust-lang/rustfmt/blob/rust-1.26.0/tests/lib.rs#L383

kolloch:
> I rebased on master and tests pass !

Awesome. Thanks for checking. @symphorien 

I'll leave this PR to @andir to finish reviewing in detail, since he wrote the test code and I still haven't looked at it in detail, unfortunately. 

My personal hunch is not to make anything writeable if there aren't problems with this. If there are IO errors related to that, I think we can provide guidance here. Generally, if it is common at all and not realistic to convince crate owners to change their code, I'd integrate this kind of things. Copy on write would be useful ;)

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -111,9 +111,30 @@ rec {
...
+          chmod +w -R .
+
+          # test harness executables are suffixed with a hash
+          # this allows to prevent name collision with the main
+          # executables of the crate
+          hash=$(basename $out)
           for file in ${drv}/tests/*; do
-            $file $testCrateFlags 2>&1 | tee -a $out
+            f=$testRoot/$(basename $file)-$hash
+            cp $file $f
```
kolloch: So I guess that copying the binaries (above) is necessary so that they are found by the test framework that you are using in https://github.com/symphorien/nix-du.

But why copy the test executables themselves? Do they really check their own path rather than the current working directory? And if yes, wouldn't a symlink surfice? Or could you delete them after a test run?

Either way, the comment should make this clear.

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -111,9 +111,30 @@ rec {
...
+
+          # the source for test data
+          cp -r ${crate.src}/* .
+
+          # build outputs
+          testRoot=target/debug
+          mkdir -p $testRoot
+
+          # executables of the crate
+          cp -r ${crate}/bin/* $testRoot
```
kolloch: Shouldn't symlinks suffice?

kolloch:
A test case (e.g. with nix-du but preferrable a more minimal test case) would be great. If it is small, then it is also easy to see what is needed.

crate2nix/templates/nix/crate2nix/default.nix
symphorien: Yes, cli_test_dir uses `std::env::current_exe()`. So a symlink would resolve to the read only one.

crate2nix/templates/nix/crate2nix/default.nix
symphorien: same reason.

symphorien:
> > This was just a hunch, but it is actually corroborated by the code:
> 
> FYI. That code does not appear to overwrite any files in the source. Instead, it looks like this test collects errors and warnings generated by `rustfmt` when checking files in its own source tree.
> 
> All calls to `format_input` (e.g. [L383 [1]](https://github.com/rust-lang/rustfmt/blob/rust-1.26.0/tests/lib.rs#L383)), pass `None` as the output stream/file, so no files appear to be changed.
> 
> If there's some other indirect function call that modifies the source tree, I don't see it upon first glance.
> 
> [1]: https://github.com/rust-lang/rustfmt/blob/rust-1.26.0/tests/lib.rs#L383

When a test fail, it does reformat the corresponding file.
I was only claiming that in this instance, not making files writable obfuscates the error message of failed tests.


~~~

I'm going to remove this from the PR, as it is so contentious. But I am slightly disappointed. This will spare milliseconds of CI but spoil hours of the next few people bumping into this issue. I don't know what is more worth caring about.

symphorien:
I removed the copy of the source.

I added a sample repo. Is there something to do to "hook" it in the test suite ?

kolloch:
> I removed the copy of the source.
> 
> I added a sample repo. Is there something to do to "hook" it in the test suite ?

You would add it to the `tests.nix`. There are already quite some samples there, you can probably adjust it.

[I should document it]

That said, @andir would still need to make a decision if he likes the general direction of this. For me, I have some other things on my list and I'd like to keep things related to test support for him to decide.

symphorien:
done

andir:
I'll play around and look at this within the next hours.

crate2nix/Cargo.nix
```diff
@@ -2746,9 +2746,31 @@ rec {
...
+          # the source for test data
+          ${pkgs.xorg.lndir}/bin/lndir ${crate.src}
+
+          # build outputs
+          testRoot=target/debug
+          mkdir -p $testRoot
+
+          # executables of the crate
+          # we copy to prevent std::env::current_exe() to resolve to a store location
+          cp -r ${crate}/bin/* $testRoot
```
andir: This fails when there are no `bin` directory in the compiled crate. We should probably handle that gracefully.

crate2nix/Cargo.nix
```diff
@@ -2746,9 +2746,31 @@ rec {
...
+
+          export RUST_BACKTRACE=1
+
+          # recreate a file hierarchy as when running tests with cargo
+
+          # the source for test data
+          ${pkgs.xorg.lndir}/bin/lndir ${crate.src}
+
+          # build outputs
+          testRoot=target/debug
```
andir: What is the reasoning for calling it `target/debug`? To mimic how cargo does it?

crate2nix/Cargo.nix
andir: This probably also means we are lacking a lib build that is being tested.

crate2nix/Cargo.nix
symphorien: yes.

crate2nix/Cargo.nix
symphorien: I replaced this with a for loop for the case where the glob expands to the empty string. I don't understand the second comment.

andir: I just meant that since tests did pass on this repository and failed on some of my projects we should probably add a test to this repo where we build a `lib` crate that is being tested.

andir:
I did give this another go but today a bunch of tests failed for weird reasons. I'll have to dig into the source of this again. This might be unrelated (hopefully).

andir:
I figured out the error from the other day. Taking another look right now. If everything checks out this should get merged today.

andir:
Thank you @symphorien :+1:

PR #128, opened 2020-04-18
direnv: fallback to use_nix if lorri is not present

Mic92:


andir:
Thanks! I actually had something similar locally :-)

PR #142, opened 2020-06-22
Add support for target_feature cfgs

### LINKED ISSUE #136: Cannot compile 'wasm-bindgen-futures' due conditional dependency on target feature
I am currently trying to build a crate that as one of its transitive dependencies `wasm-bindgen-futures` and from this code: https://github.com/rustwasm/wasm-bindgen/blob/master/crates/futures/Cargo.toml#L18 it generates the following code:

```
target = { target, features }: (target."feature" == "atomics");
```

This then errors with: `attribute 'feature' missing`.

It is similar to https://github.com/kolloch/crate2nix/issues/37 but it uses `target_feature` here instead of just `target`

kristoff3r:
This PR adds support for [target-feature](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-feature) cfgs, by allowing users to specify the target features manually. As the [gotchas page](https://doc.rust-lang.org/rustc/targets/known-issues.html) notes this is potentially unsafe, so I'm very open to suggestions on more automated / safer ways to do this. 

My use case is a project that depends on the `sha` crate, which runs fine when I supply `[ "sha" "sse2" ]` as targetFeatures.

This fixes #136.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -14,6 +14,8 @@
 , rootFeatures ? [ "default" ]
   # If true, throw errors instead of issueing deprecation warnings.
 , strictDeprecation ? false
+  # Used for conditional compilation based on CPU feature detection.
+, targetFeatures ? []
   # Whether to perform release builds: longer compile times, faster binaries.
 , release ? true
   # Additional crate2nix configuration if it exists.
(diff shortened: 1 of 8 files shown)


danieldk:
Thank you! I have been using this in one of my projects where I hit the same issue as #136 and it works like a charm!

kolloch:
merged

PR #143, opened 2020-07-08
Avoid duplicate test builds

andir:
We will now just create a derivation that symlinks to the original outputs instead of repeating the entire build just for ensuring the tests have been run. Depending on the complexity (and size) of the builds duplicating the effort might be rather expensive.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -144,17 +144,15 @@ rec {
           done
         '';
     in
-    crate.overrideAttrs
-      (
-        old: {
-          checkPhase = ''
-            test -e ${test}
-          '';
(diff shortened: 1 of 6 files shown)


crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -144,17 +144,15 @@ rec {
           done
```
kolloch, line 144: Awesome opportunity to check up on the comments -- do they make sense if we don't have all of the context in our brains anymore?

I cannot comment the function doc directly because it is not close enough to the change. It says "Returns a crate" where it probably meant to say "Returns a derivation"

It would also be nice to document "crate" vs "testCrate"

kolloch:
merged

PR #151, opened 2020-09-20
[WIP] callCrate: IFD function to call use crate2nix without Cargo.nix

### LINKED ISSUE #148: Production use (non-local)
Obviously, forking every repo to add a Cargo.nix file is absurd, and nixpkgs never took that approach. Rather, the source is treated as unalterable, patched in the worst case, and given as an argument to a builder. But after looking through the source and GH issues, there is no indication that it is possible to use crate2nix this way.
`crate2nix generate -f https://github.com/kolloch/crate2nix -o ~/wrapperNixFlake/cargo.nix` ??

balsoft:
## Usage (all examples work fine)

crate2nix itself:

```
nix-build -E '((import (builtins.fetchTarball https://github.com/balsoft/crate2nix/archive/balsoft/callcrate.tar.gz) {}).callCrate "${builtins.fetchTarball https://github.com/kolloch/crate2nix/archive/master.tar.gz}/crate2nix" {}).rootCrate.build' --show-trace
```

cargo-local-registry, which was very useful in figuring out the format for local registries:

```
nix-build -E "((import (builtins.fetchTarball https://github.com/balsoft/crate2nix/archive/balsoft/callcrate.tar.gz) {}).callCrate (builtins.fetchTarball https://github.com/ChrisGreenaway/cargo-local-registry/archive/master.tar.gz)).rootCrate.build" --show-trace
```

Lemmy, quite a big one that works out of the box:

```
nix-build -E "((import (builtins.fetchTarball https://github.com/balsoft/crate2nix/archive/balsoft/callcrate.tar.gz) {}).callCrate (builtins.fetchTarball https://github.com/LemmyNet/lemmy/archive/main.tar.gz)).workspaceMembers.lemmy_server.build" --show-trace
```

A more complex example with overrides, workspace and flakes: https://github.com/balsoft/simple-osd-daemons/commit/35a17a8862139a3b52e2b729dbea1d1b319b41ed

## Related issues

Closes #148 

Initial idea in #8

## Pings

cc @maisiliymBridj @xcthulhu

## Further work

- FIXME: Probably needs a flag on the Rust side of things;
- FIXME: Copying the entirety of crates.io-index into our local repository is a waste;
- TODO: Update documentation and add examples;
- TODO: Probably refactor a bit.

## Questions

Q: Why does it need to fetch all the dependencies at eval-time?

A: Because `cargo metadata` requires this

***

Q: Why does `cargo metadata` require all the dependencies?

A: To calculate features, among other things


callCrate.nix
```diff
@@ -0,0 +1,61 @@
...
+    mkdir $out
+    cp ${crates-io-index} -r $out/index
+  '' + lib.concatMapStringsSep "\n" (crate: "cp ${crate} $out/${builtins.replaceStrings [".tar.gz"] [".crate"] crate.name}") crates);
+
+  cargoConfig = writeText "config.toml" ''
+    [source.crates-io]
+    replace-with = "nix-crates-registry"
+
+    [source.nix-crates-registry]
+    local-registry = "${source-repository}"
```
xcthulhu: Won't we also want:

```toml
[net]
offline = false
```
?

https://doc.rust-lang.org/cargo/reference/config.html#netoffline

balsoft: I think we also need a way to add extra registries here. I'll probably do that next weekend.

xcthulhu: I think it will error our either way, but it might error out more helpfully this way...

balsoft: Actually, now that I think about it, crate2nix itself doesn't support non-crates.io registries yet, so it doesn't make sense to support them here. I'll just add `offline = true`.

balsoft: Done.

xcthulhu:
I'm trying to test this out... I am not sure I am using this correctly.

Here's what I am trying:

```bash
cd /tmp
cargo new my-test
#     Created binary (application) `my-test` package

cd my-test/                                                                                                                      
cargo generate-lockfile

cat > default.nix <<EOF                                                                                                  
let                                                                                                                                                
  crate2nix = import (builtins.fetchGit {                                                                                                          
    url = "https://github.com/balsoft/crate2nix/";                                                                                                 
    ref = "balsoft/callcrate";                                                                                                                     
    rev = "5c01f5daac71125122277cdbd6e225e6405629b3";                                                                                              
  }) { };                                                                                                                                          
  project = crate2nix.callCrate {                                                                                                                  
    src = builtins.fetchTarball                                                                                                                    
      "https://github.com/kolloch/crate2nix/archive/master.tar.gz";                                                                                
  };                                                                                                                                               
in project.rootCrate.build                                                                                                                         
EOF 

nix-shell                                                                                                                
#error: opening file '/nix/store/j80gkkd8bq8vwv6vm9362vlmjqqv9adq-source/Cargo.lock': No such file or directory                                     
(use '--show-trace' to show detailed location information)
```

balsoft:
Oh, sorry, I forgot that it needs  `/crate2nix` subfolder... Fixed now.

xcthulhu:
Hmm... still getting that even on https://github.com/kolloch/crate2nix/pull/151/commits/4e95760a4a1eb8394bf6610b457f610f7cb033db .

In my example I should have `src = ./.` rather than `src = builtins.fetchTarball "https://github.com/kolloch/crate2nix/archive/master.tar.gz"`.

When I switch it to `./.` it appears to work (still compiling...).

But I think it should also work with `src = builtins.fetchTarball "https://github.com/kolloch/crate2nix/archive/master.tar.gz"`?

balsoft:
I have fixed the example in original comment, it now reads
```
let
  inherit (import sources.crate2nix { }) callCrate;
  project = callCrate { src = "${builtins.fetchTarball https://github.com/kolloch/crate2nix/archive/master.tar.gz}/crate2nix"; };
in project.rootCrate.build
```

The issue was that the crate is located in `crate2nix` subfolder.

balsoft:
Changed all examples to be easily executable.

crate2nix.nix
```diff
@@ -0,0 +1,66 @@
...
+        name = "Peter Kolloch";
+      }
+      lib.maintainers.andir
+    ];
+    platforms = lib.platforms.all;
+  };
+  postBuild = ''
+    # Fallback to built dependencies for cargo and nix-prefetch-url
+    wrapProgram $out/bin/crate2nix ${set_templates}\
+    --suffix PATH ":" ${lib.makeBinPath [ cargo nix nix-prefetch-git ]}
```
xcthulhu, line 58: CI is crashing with `Builder called die: makeWrapper doesn't understand the arg PATH`

I see this is the same as the old `default.nix`, so I'm not sure how this passed CI before...

https://github.com/kolloch/crate2nix/pull/151/checks?check_run_id=1144007710#step:5:550

kolloch:
Hi @balsoft , isn't that very similar to what https://github.com/kolloch/crate2nix/blob/master/tools.nix does?

xcthulhu:
@kolloch Hmm... `tools.nix` looks pretty good!  I see here's how we'd use it like @balsoft 's `callCrate`:

```nix
{ pkgs ? import <nixpkgs> { } }:
let
  crate2nix-tools = import "${
      builtins.fetchGit {
        url = "git@github.com:kolloch/crate2nix.git";
        ref = "master";
        rev = "e07af104b8e41d1cd7e41dc7ac3fdcdf4953efae";
      }
    }/tools.nix" { };
  project = pkgs.callPackage (crate2nix-tools.generatedCargoNix {
    name = "websocat";
    src = builtins.fetchTarball
      "https://github.com/vi/websocat/archive/master.tar.gz";
  }) { };
in project.rootCrate.build
```

In @balsoft 's code he parses `Cargo.toml` to figure out the derivation name... maybe that would be a good idea for `generatedCargoNix`?

balsoft:
Ok, so I clearly should have looked better :P I think this PR should be repurposed to automatically parse the crate name in `tools.nix` ? And maybe also in the next PR I'll add a `flake.nix` here for easier use with flakes.

PR #154, opened 2020-10-25
tools.nix Nix 3.0 improvements

balsoft:
- Nix 3.0 has `submodules` argument of `builtins.fetchGit`, which allows us to fetch `git` dependencies without `crate-hashes.json`
- Nix 3.0 introduces flakes, which force restricted evaluation mode. Make some changes to `src` logic so that restricted evaluation succeeds.


kolloch:
Sorry, for letting you wait that long. Looks cool!

PR #157, opened 2020-11-28
Handle situations where there are non-feature cfgs other than well-known ones

### LINKED ISSUE #141: Non-existent feature references in `target.cfg(...)` will cause evaluation failures
It is not invalid to have a crate like this:

```toml
# cat Cargo.toml
[package]
name = "wow"
version = "0.1.0"

# Effectively a no-op, however cargo still verifies that the dependency can be resolved
[target.'cfg(banana)'.dependencies]
psm = "0"
```

But when generated with crate2nix and built it will fail:

```
$ crate2nix generate -o Cargo.nix
Generated Cargo.nix successfully.
$ nix build -f Cargo.nix workspaceMembers.wow
error: attribute 'banana' missing, at /tmp/wow/Cargo.nix:117:44
(use '--show-trace' to show detailed location information)
```

A real project with such a problem is [tokio-bytes](https://github.com/tokio-rs/bytes/blob/008d3e508e2247dbfbf5a73918e2fe999a4e9158/Cargo.toml#L33).

cc @jlas

nagisa:
Fixes #141


kolloch:
Thx, @nagisa !

PR #159, opened 2020-12-24
tools.nix: compare Nix version with 2.3

balsoft:
Previously, the new Nix release would be called 3.0. Now the version
is 2.4pre, which means that matching the major version with 3 to
determine if new features are available is incorrect.

Tested on balsoft/simple-osd-daemons, without this patch it fails to fetch `mpris` dependency via git and with this patch everything builds correctly.

### DIFF
--- a/tools.nix
+++ b/tools.nix
@@ -245,8 +245,9 @@ rec {
             value = builtins.readFile hash;
           };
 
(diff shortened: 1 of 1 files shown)


kolloch:
Thx, @balsoft !!

PR #162, opened 2021-02-06
fix test regression

pandaman64:
Fixed a unit test regressed by #157.
This regression seems to slip through the CI due to [`--no-cargo-build`](https://github.com/kolloch/crate2nix/blob/509918492f7c5e6cff1697e1857a383f46bf6a76/.github/workflows/tests-nix-linux.yml#L21).


kolloch:
Thank you so much, @pandaman64 !

PR #166, opened 2021-02-19
WIP: add support for veloren in tools.nix

marius851000:
This is my WIP work to be able to automatically generate the Cargo.nix for veloren via the tools.nix. The following change were made:

- allow fetching workspace git depencies
- allow to download git dependancies without an sha256 (that was work I did a long time ago, and don't remember exactly how it work, but it does)

What remain to do:

- figure why num-literal is compiled without num-bigint, so failing the compilation of num that expect num-literal to be compiled with num-bigint (if you can help with that. That seems to be an issue with dynamic feature resolution).

### DIFF
--- a/crate2nix/src/resolve.rs
+++ b/crate2nix/src/resolve.rs
@@ -375,7 +375,7 @@ impl From<crate::config::Source> for ResolvedSource {
                 url,
                 rev,
                 r#ref: None,
-                sha256: Some(sha256),
+                sha256: sha256,
             }),
             crate::config::Source::CratesIo {
                 name,
--- a/crate2nix/src/config.rs
+++ b/crate2nix/src/config.rs
@@ -105,7 +105,7 @@ pub enum Source {
         /// The revision hash.
         rev: String,
         /// The sha256 of the fetched result.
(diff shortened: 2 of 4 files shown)


kolloch:
Heyo @marius851000,

I am, unfortunately, time constrained.

So, can you pleasse elaborate? What is the intent of each change? How does it work?

E.g. if you allow git references without sha, what is supposed to happen? Prefetch, determine sha? When is it supposed to happen?

What are workspace git dependencies? I see some string constants in the code and it is hard for me to judge if they are right.

Maybe also the following would help: Add any usage changes to README.md. Add code comments to explain things.

crate2nix/src/resolve.rs
```diff
@@ -375,7 +375,7 @@ impl From<crate::config::Source> for ResolvedSource {
                 url,
                 rev,
                 r#ref: None,
-                sha256: Some(sha256),
+                sha256: sha256,
```
zseri, line 378:
```suggestion
                sha256,
```

kolloch:
Heyo, I am closing this for now, I hope that's fine.

PR #160, opened 2021-01-07
Add support for cross-compilation, with backcompat

Ericson2314:
In https://github.com/kolloch/crate2nix/pull/152#issuecomment-754203498 I made a PR against a PR, but i get not CI that way, so I'm opening this too.

PR #178, opened 2021-04-04
document cross support

symphorien:
fix readme to not recommend non-cross-friendly constructs that use
deprecrated buildRustCrate argument to Cargo.nix
improve warning
fix this warning in default.nix

Source: https://github.com/kolloch/crate2nix/pull/160/files#diff-381a73e3c364568cfbd7c56e32e1db595235606948c51c06f04ff99d75aec589

cc @Ericson2314 

Motivation: this warning mentions buildRustCrate but in nix-du's nix files there is no occurence of buildRustCrate in manually written files. It took me really long to understand why the warning was fired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Within a nix file (e.g. your manually written `default.nix`), you can access the
 derivation like this:
 
 ```nix
-let cargo_nix = callPackage ./Cargo.nix {};
+let cargo_nix = import ./Cargo.nix { inherit pkgs; };
 in cargo_nix.rootCrate.build
 ```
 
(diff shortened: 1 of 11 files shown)


symphorien:
The ci is flakey, I give up.

symphorien:
ping @kolloch

Ericson2314:
Is getting rid of `callPackage` related?

zseri:
this should be squashed at merge.

@Ericson2314 yes.

kolloch:
Thank you very much :)

Feel free to add yourself to the changelog.

Ericson2314:
It would be nice if `callPackage` was still OK though.

kolloch:
@Ericson2314 I agree. Basically, I strive to prevent breaking anything that worked before, except it is clearly marked as internal. Is it broken now?

I missed the change that breaks it.

symphorien:
you can `callPackage { buildRustCrate = null; }` but that's verbose imo

Issue #182, opened 2021-04-09
Can't build crate with dependency num 0.4.0


1000teslas:
I get this error
```
error[E0432]: unresolved import `num_rational::BigRational`
  --> src/lib.rs:68:9
   |
68 | pub use num_rational::BigRational;
   |         ^^^^^^^^^^^^^^-----------
   |         |             |
   |         |             help: a similar name exists in the module: `Rational`
   |         no `BigRational` in the root
```
when trying to `nix-build` the generated `Cargo.nix`. However `cargo build` works fine.

pandaman64:
Dupe of #146

kolloch:
Hey @1000teslas,

thank you so much for reporting this! 

@pandaman64 has fixed this in a recent PR. Would you help by retesting it with the master version?

Thank you.

1000teslas:
seems like it works fine. thanks.

kolloch:
Awesome, thanks for getting back to us!

bbigras:
I had the same problem and master also fixed it. Thanks for the quick fix!

PR #185, opened 2021-04-22
Adds optional dependencies to features when they are enabled indirectly

### LINKED ISSUE #146: Features fail to contain optional dependencies that got enabled indirectly
Consider the following Cargo.toml:

```toml
[package]
name = "test-num-features"
version = "0.1.0"
edition = "2018"

[dependencies]
num = { version = "=0.3", default-features = false, features = ["std"] }
```

It currently fails to build despite succeeding with cargo. The reason is because when `crate2nix` is building `num-rational` it somehow drops the feature for an optional `num-bigint` dependency. This optional dependency ends up being enabled indirectly, through `num` depending on both `num-rational` and `num-bigint`.

Both `cargo metadata` and the generated `Cargo.nix` contains `num-bigint` in the `resolvedDefaultFeatures`:

```nix
{
      "num-rational" = rec {
        crateName = "num-rational";
        version = "0.3.0";
        edition = "2018";
        sha256 = "0f41j1l1kn5jj36a8xdy8kv242wlwq0ka578vm8gnb1n1wvdgd55";
        authors = [<snip>];
        dependencies = [<snip>];
        buildDependencies = [<snip>];
        features = {
          "default" = [ "num-bigint-std" "std" ];
          "num-bigint-std" = [ "num-bigint/std" ];
          "std" = [ "num-integer/std" "num-traits/std" ];
        };
        resolvedDefaultFeatures = [ "num-bigint" "num-bigint-std" "std" ];
      };
}
```

But the final invocation of `buildRustCrate` ends up not being provided the feature in the end. And so `num-rational` is compiled without a `--cfg feature="num-bigint"` flag.

I admit that this is the first time I’m reading [templates/nix/crate2nix/default.nix](https://github.com/kolloch/crate2nix/blob/master/crate2nix/templates/nix/crate2nix/default.nix) at all seriously, but from what I can tell it just throws away the information `cargo` already computed and attempts to compute its own feature sets. Is that correct? Why?

pandaman64:
When Cargo sees a dependency feature `opt_crate/feat` for an optional dependency `opt_crate`,
it implicitly adds `opt_crate` to the feature set for the current crate.
For example, `num-rational` crate relies on this behavior for importing `num-bigint`, which is optional.
This patch lets crate2nix mimic this behavior.

Main changes are in [crate2nix/templates/nix/crate2nix/default.nix](https://github.com/kolloch/crate2nix/compare/master...pandaman64:add-optional-dep-as-feature?expand=1#diff-573bf3b71d0d8bf3ff7c415529cde20e362cf673c32b75d4fe170ed880c39655)
I also modified test cases([crate2nix/templates/nix/crate2nix/tests/packageFeatures.nix](https://github.com/kolloch/crate2nix/compare/master...pandaman64:add-optional-dep-as-feature?expand=1#diff-ae3d76bc49319d02ff1dd589344d2b695cd301cdee52cc85250ab9c4eb91ecc6)) so that they match what Cargo does.

Closes #146, #182.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -2596,10 +2596,11 @@ rec {
       let
         crateConfig = crateConfigs."${packageId}" or (builtins.throw "Package not found: ${packageId}");
         expandedFeatures = expandFeatures (crateConfig.features or { }) features;
+        enabledFeatures = enableFeatures (crateConfig.dependencies or [ ]) expandedFeatures;
         depWithResolvedFeatures = dependency:
           let
             packageId = dependency.packageId;
-            features = dependencyFeatures expandedFeatures dependency;
+            features = dependencyFeatures enabledFeatures dependency;
           in
           { inherit packageId features; };
         resolveDependencies = cache: path: dependencies:
@@ -2608,7 +2609,7 @@ rec {
           let
             enabledDependencies = filterEnabledDependencies {
               inherit dependencies target;
-              features = expandedFeatures;
+              features = enabledFeatures;
             };
             directDependencies = map depWithResolvedFeatures enabledDependencies;
             foldOverCache = op: lib.foldl op cache directDependencies;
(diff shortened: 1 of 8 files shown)


kolloch:
Woah, thank you, @pandaman64!

Prime example of a PR I love.

1. You clearly write what you want to achieve and what you changed. (that significantly reduces the time that I need to review it)
2. You have nice inline comment that also explains it.
3. You adjusted the unit tests.

:fireworks:

Fuuzetsu:
Thanks! Just confirming that it works for us too, I tried it today.

PR #193, opened 2021-05-05
Allow arbitrary commands before and after tests

### LINKED ISSUE #191: Allow test crate to fail but still "build"
We are running tests in multiple crates, creating the test report with something like `cargo2junit` and showing it in the CI. But if the crate just fails, we can't see the (failing) output and handle the failure reporting ourselves.

Would be cool to unconditionally produce test output even if it failed.

Behind some flag of course.

Fuuzetsu:
An immediate use-case is to inject any environment variables that the
test may be relying on, such as location of some external files
or overriding some behaviour.

Another, perhaps more cliche use-case is for unsetting the `-e` bash
flag: this will allow us to capture output of failing tests without
giving up on expression itself. In this way, it fixes #191.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -96,9 +96,11 @@ rec {
      testCrateFlags: list of flags to pass to the test exectuable
      testInputs: list of packages that should be available during test execution
   */
-  crateWithTest = { crate, testCrate, testCrateFlags, testInputs }:
+  crateWithTest = { crate, testCrate, testCrateFlags, testInputs, testPreRun, testPostRun }:
     assert builtins.typeOf testCrateFlags == "list";
     assert builtins.typeOf testInputs == "list";
+    assert builtins.typeOf testPreRun == "string";
+    assert builtins.typeOf testPostRun == "string";
     let
       # override the `crate` so that it will build and execute tests instead of
       # building the actual lib and bin targets We just have to pass `--test`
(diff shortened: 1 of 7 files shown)


Fuuzetsu:
I verified that I can deal with #191 just by doing

```
testPreRun = "set +e";
testPostRun = "set -e";
```

I also verified that when neither of these flags is set, I don't have any rebuild of the tests so any existing users should not be affected by the change.

Fuuzetsu:
@kolloch Any updates? It'd be cool if we could use the repo directly rather than our fork. We've been using this change for a while with success.

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -112,6 +114,15 @@ rec {
                 buildTests = true;
               }
             );
+          # If the user hasn't any pre/post commands, we don't want to insert
+          # empty lines. This means that any existing users of crate2nix don't
+          # get a spurious rebuild unless they set these.
```
kolloch, line 119: Thank you for being that diligent!

kolloch (changes requested):
This looks simple and good. Thank you.

Could you document this in the README.md please and regenarate generated sources by executing `run_tests.sh`? (sorry, this will download the internet)

Fuuzetsu:
>Could you document this in the README.md please and regenarate generated sources by executing run_tests.sh? (sorry, this will download the internet)

Done. All tests passed apparently. I committed the changes. I fixed small typo/grammar issue in the comment, nothing major.

Feel free to change the README.md sections as you see fit if you imagined something a bit different.

kolloch (approved):
Thank you :)

PR #202, opened 2021-07-07
Bump deps, use cargo_platform to avoid vendored code

Ericson2314:
The git dep is used for `cargo_metadata` to align the `semver` versions. https://github.com/oli-obk/cargo_metadata/issues/164 tracks that and it is supposed to be resolved soon.


Ericson2314:
OK Hmm I am wrong. It works when run it in crate2nix's nix shell, it doesn't. After more testing, the difference still holds even if I clear the path in the shell and use the nix-built version rather than `cargo run`. Weird!

Ericson2314:
@nagisa if you are curious, here is where I have been bumbling about that issue.

PR #203, opened 2021-07-08
Filter platform-specific dependencies according to *local* info

Ericson2314:
Now that we support cross compilation, it's wrong to have a global
`target`. Instead, we keep per-platform metadata for `cfg` the same way
we keep per-platform derivations.


Ericson2314:
I'm happy to report we finally have a cross test!

PR #206, opened 2021-07-16
Allow users to configure default git branch name

JonathanLorimer:


Ericson2314:
@QAston sorry I didn't respond sooner, are you saying this fix is insufficient on its own?

QAston:
Yes, I think so. Might be a different problem though - my cargo file pointed to the correct branch but it was ignored by cargo.nix, maybe in the OP's case there was no branch set in cargo files.

kolloch:
Heyo @QAston / @Ericson2314,

I am doing some clean up work... I guess we can close this?

kolloch:
(otherwise reopen)

QAston:
no idea, I moved on from nix, I hope I never have to go back

PR #212, opened 2021-10-05
Use the rust-specific config

Ericson2314:
Per the TODO, we could leverage Nixpkgs to do this better still, but it
isn't terribly convenient at this time. I rather do it after the other
cross PRs are merged.


Ericson2314:
I'm not longer worried about testing this so much, because now that it uses a function and isn't conditional, the ball is more in Nixpkgs court (there is a single code path on the create2nix side).

So merely ensuring that existing functionality still works should be good.

PR #214, opened 2021-10-25
refactor(github/workflows/niv-updater): run once a week

lovesegfault:
The current updater workflow is very hyper, polluting the commit logs with a
flood of update commits.

This changes it to only run once a week, which hopefully is more than enough.


Ericson2314:
The old one seemed to have stopped anyways, so we might as well try this.

Issue #213, opened 2021-10-22
IFDs break with newer Nix


lorenzleutgeb:
See upstream issue https://github.com/NixOS/nix/issues/5410

balsoft:
Should be fixed now

PR #215, opened 2021-11-03
Fix broken IFD build with Nix 2.4

balsoft:
Because of https://github.com/NixOS/nix/issues/5410, it's not possible
to filter paths with references anymore. This prevented any IFD-based
crate2nix usage. Fix this by not filtering the source when the Nix
version used includes the breaking commit.

Works around #213

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -454,7 +454,7 @@ rec {
         ];
         # We can't filter paths with references in Nix 2.4
         # See https://github.com/NixOS/nix/issues/5410
-        src = if (lib.versionOlder "2.4" builtins.nixVersion)
+        src = if (lib.versionOlder "2.4pre20211007" builtins.nixVersion)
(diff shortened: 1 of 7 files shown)


Ericson2314:
@balsoft Any idea why the tests are still failing?

balsoft:
That's interesting

lorenzleutgeb:
Could you please explain whether/how this relates to https://github.com/NixOS/nix/pull/5494? Thank you!

Ericson2314:
Because of https://github.com/NixOS/nix/pull/5494 Nix 2.5 should just work and we shouldn't need this workaround no more.

kolloch:
Heyo, I merged this. Thank you @lorenzleutgeb ! 

Even thought this should not be necessary for long, I do like working tests meanwhile :)

symphorien:
I see the issue is closed, could this workaround be removed for fixed nix versions ? I am using crate2nix without IFD and this causes the whole target directory to be copied...

PR #216, opened 2021-11-04
Fix logic for enabling features based on renamed dependencies

duxovni:
When `enableFeatures` inspects optional dependencies to enable additional features in accordance with https://doc.rust-lang.org/nightly/cargo/reference/features.html#dependency-features , it can incorrectly enable features for the original names of renamed dependencies, and fail to enable features for the new names. For example, diesel has a feature `"uuid"` which adds `uuid < 0.7.0` as a dependency, and a separate feature `"uuidv07"` which adds `uuid >= 0.7.0` renamed as `uuidv07`. When the `"uuidv07"` feature is enabled, `enableFeatures` sees that one of the features enables the dependency `uuid`, so it also enables the `"uuid"` feature. However, the relevant dependency was renamed, and the non-renamed `uuid < 0.7.0` required by the `"uuid"` feature isn't present among the dependencies, so the build breaks.  This patch fixes `enableFeatures` and `doesFeatureEnableDependency` to correctly handle renamed dependencies.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -2818,15 +2818,14 @@ rec {
       dependencies;
 
   /* Returns whether the given feature should enable the given dependency. */
-  doesFeatureEnableDependency = { name, rename ? null, ... }: feature:
+  doesFeatureEnableDependency = dependency: feature:
     let
+      name = dependency.rename or dependency.name;
       prefix = "${name}/";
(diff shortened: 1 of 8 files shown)


Ericson2314:
Thanks for this!

Ericson2314:
I tested this by hand, thanks!

PR #219, opened 2021-11-07
Update cachix actions

Ericson2314:
Hopefully this starts to unbreak the build with Nix 2.4


Ericson2314:
Makes it less broken at least.

PR #224, opened 2022-05-05
Support package renaming inside dev-dependencies

### LINKED ISSUE #223: Renames outside of `[dependencies]` sections aren't tagged as such
If you have something like
```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]

[dev-dependencies]
renamed_itoa = { version = "0.4.8", features = ["i128"], package = "itoa" }
```

crate2nix does not generate a renamed package, it instead generates something like
```nix
{
            name = "itoa";
            packageId = "itoa 0.4.8";
            features = [ "i128" ];
}
```

It does the right thing if the dependency is moved to `[dependencies]`

```nix
          {
            name = "itoa";
            packageId = "itoa 0.4.8";
            rename = "renamed_itoa";
            features = [ "i128" ];
          }
```

It seems like this is intended to work as I see this:

```nix
dependenciesWithRenames =
              lib.filter (d: d ? "rename")
                (
                  filterEnabledDependenciesForThis
                    (
                      (crateConfig.buildDependencies or [ ])
                      ++ (crateConfig.dependencies or [ ])
                      ++ devDependencies
                    )
                );
```

Maybe somewhere in crate2nix the rename code forgot to be plugged in for `dev-dependencies` and `build-dependencies`?

This is a problem if we have overlapping but renamed packages outside of `dependencies`.

Fuuzetsu:
It already works for `dependencies` and `build-dependencies`, just seems
like an oversight.

I ran `cargo test`, seems to pass.

Fixes #223.


Ericson2314:
Thanks!

Would you mind adding a test?

Fuuzetsu:
I'm not sure how. I changed `sample_projects/renamed_build_deps/Cargo.toml` to include something like

```diff
diff --git a/sample_projects/renamed_build_deps/Cargo.toml b/sample_projects/renamed_build_deps/Cargo.toml
index 8c26807..e49ffa7 100644
--- a/sample_projects/renamed_build_deps/Cargo.toml
+++ b/sample_projects/renamed_build_deps/Cargo.toml
@@ -10,3 +10,7 @@ edition = "2018"
 
 [build-dependencies]
 which_crate = {package = "which", version = "3.0", default-features = false}
+
+# Checks that renames in dev-dependencies work.
+[dev-dependencies]
+futures01 = { package = "futures", version = "0.1", default-features = false }
```

I ran `cargo check` to update the lock file and then I ran `./run_tests.sh`. It just told me that everything looks OK except uncommitted changes. Cargo.nix doesn't seem updated there. Running `./regenerate_cargo_nix.sh` explicitly doesn't seem to help either.

Any suggestion? Or if it's faster for you to add one, feel free...

Ericson2314:
It took some digging, but the file that confusingly wasn't changing is just dead code! https://github.com/kolloch/crate2nix/pull/245

Ericson2314:
I am making the tests needed, so don't worry about it :)

PR #250, opened 2022-08-08
Try pinning the macOS version in github action base image

Ericson2314:


Ericson2314:
There are other issues, but it definitely got further.

PR #254, opened 2022-08-23
expose required-features to buildRustCrate

marius851000:
The required features field allow to skip the compilation of a binary (or test, bench and example) if all the features aren’t enabled. This is documented here : https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field

This PR against nixpkgs make use of it: https://github.com/NixOS/nixpkgs/pull/187975

It should still work independantly of whether this nixpkgs PR is merged or not.


sample_projects/bin_required_features/src/bin/some_binary_with_rand.rs
```diff
@@ -0,0 +1,7 @@
+use rand::prelude::*;
+
+pub fn main() {
+    let mut rng = rand::thread_rng();
+    let y: f64 = rng.gen();
+    println!("The binary that should NOT be compiled found the random number {}", y);
```
Ericson2314, line 6: Let's instead use https://doc.rust-lang.org/std/macro.compile_error.html to make building this package and error so the test is more stringent?

I am happy to merge the Nixpkgs PR in a second per the comment I left, so don't worry that doing the above will make CI fail today.

sample_projects/bin_required_features/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "bin_required_features"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rand = { version = "0.8.5", optional = true }
+
```
Ericson2314, line 10: Can you give the "main" binary a required feature too? Make it one that we *do* pass so we continue building the main binary.

Thus the test would have:

1. Binary with feature we *do pass, binary can build, build is attempted.
2. Binary with feature we *don't* pass, binary can't build, build isn't attempted.

marius851000:
That's done. The newly added test passes locally (while some other fail due to the lack of ``unsafeDiscardStringContext``, fixed in my other PR). As a bonus, it remove the dependency on rand.

Ericson2314:
Thanks!

PR #255, opened 2022-08-26
Don't hard-code a target family of Unix

Ericson2314:
Rust target specs actually support multiple target families, where just
one needs to match, so we do too.

Reverse the order of the comparison so it is more consistent: wanted
before the target's single value or list.


Ericson2314:
The new logic should be backported into a `toTargetFamilies` next to the others in the "rust lib". Then this is ready to be merged.

Ericson2314:
(The tests passed locally)

PR #256, opened 2022-08-26
Delete `enum SourceType`, which is dead code

Ericson2314:
@kolloch I will leave this one to you, in case you had meant to use this for something and it would be better to write that something instead of deleting this.


Ericson2314:
I'll just merge this. It is still in history.

PR #265, opened 2022-11-25
Honor `foo?/bar` feature selections

tilpner:
This PR adds support for the new-ish `foo?/bar` feature selection syntax (1.60+, [documentation][0]). The `time` crate wasn't compiling anymore after upgrading to version 0.3.17, and this change fixes compilation in the one project I've tested this with.

[0]: https://doc.rust-lang.org/cargo/reference/features.html#dependency-features


Ericson2314:
Looking good, but can you add a test for this?

tilpner:
I've added something, but it should check that the `optional` crate is not being built in the ["foo", not "hello"] case, instead of just printing something different (not sure how to fit that into the testing setup).

tilpner:
ahh, good idea, that should work

Ericson2314:
Thanks so much! The last thing is adding a new change log entry. You can do it in a new 0.11 -- 0.12 section because I just made a release candidate for 0.11.

tilpner:
Does this look right? It looks a bit wrong, starting with two (not released) sections

Ericson2314 (approved):

lblasc:
you saved my day :) thanks!

PR #272, opened 2023-01-21
undo https://github.com/kolloch/crate2nix/pull/215 for nix >= 2.5

symphorien:
the fix is in nix >= 2.5
https://github.com/NixOS/nix/commit/67179472dfe30c97af9a5aeed0190c0e765936cf

cc @balsoft

PR #275, opened 2023-02-16
Use rust lib to set target vendor

jordanisaacs:
Uses the new rust lib function `toTargetVendor` for setting the vendor for feature/dependency resolution & regenerated `Cargo.nix` files. Needs nixpkgs to be updated to a version of 21.11 that includes `toTargetVendor`, unsure how to do that.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -34,12 +34,12 @@ rec {
     os = pkgs.rust.lib.toTargetOs platform;
     arch = pkgs.rust.lib.toTargetArch platform;
     family = pkgs.rust.lib.toTargetFamily platform;
+    vendor = pkgs.rust.lib.toTargetVendor platform;
     env = "gnu";
     endian =
       if platform.parsed.cpu.significantByte.name == "littleEndian"
       then "little" else "big";
     pointer_width = toString platform.parsed.cpu.bits;
-    vendor = platform.parsed.vendor.name;
     debug_assertions = false;
   };
 
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -2367,12 +2367,12 @@ rec {
     os = pkgs.rust.lib.toTargetOs platform;
     arch = pkgs.rust.lib.toTargetArch platform;
     family = pkgs.rust.lib.toTargetFamily platform;
+    vendor = pkgs.rust.lib.toTargetVendor platform;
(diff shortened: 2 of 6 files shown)


Ericson2314:
@jordanisaacs it is done with Niv, or you can just manually modify the file.

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -39,7 +39,7 @@ rec {
       if platform.parsed.cpu.significantByte.name == "littleEndian"
       then "little" else "big";
     pointer_width = toString platform.parsed.cpu.bits;
-    vendor = platform.parsed.vendor.name;
+    vendor = pkgs.rust.lib.toTargetVendor platform;
```
Ericson2314, line 42: I would move this up to be next to the others.

jordanisaacs:
I couldn't run the tests locally because I got this error
```
error: the string '2mz4bcqmqrkndkyqnbr57djd9fs6zqsh-crates-io-aho-corasick-0.7.18' is not allowed to refer to a store path (such as '!out!/nix/store/lpa1wknhyyin0hxd9scck1si0459hfpb-crates-io-aho-corasick-0.7.18.drv')
```

Ericson2314:
I suspect https://github.com/kolloch/crate2nix/pull/272 is not quite right. Try reverting that.

jordanisaacs:
@Ericson2314 I tried reverting it but did not change the error. Not entirely sure what is causing the error but it occurs when entering the `nix-shell --pure shell.nix`. I was runnning nix-shell on version `2.13.2`

Ericson2314:
@jordanisaacs Can you do it again with `--show-trace`?

Ericson2314:
Ah, I can reproduce.

jordanisaacs:
Here is the super long trace if you still want it: https://gist.github.com/jordanisaacs/18fa608cb59dd924ec8567077ee01ff8

jordanisaacs:
Occurs in building cargo-release

Ericson2314:
Ah I was surprised because this looked like a bug I had fix. But it was! See #277 which bumps the prior version of crate2nix pinned in crate2nix.

jordanisaacs:
@Ericson2314 Should be working now! ran `./run_tests.sh` and they all passed.

kolloch:
Heyo @jordanisaacs,

I cherry-picked your commit and rebased it on master. Thank you!

kolloch:
Thanks :)

PR #276, opened 2023-02-22
add ability to override lockfile from tools.nix

apoelstra:
I would like the ability to override the Cargo.lock file in a repository (or provide one if none exists). It looks like there is no way to do this through Cargo itself -- its source code hardcodes the path of the lockfile as the workspace root plus `/Cargo.lock`. So to override a lockfile you have to copy it in place, which is what this patch does.

I'm very new to nix, so any feedback on approach is welcome.

My use case is that I'm using this crate as part of a CI infrastructure, where for a given pull request I grab every commit, dynamically produce a Cargo.nix, and build them. This is a library so it doesn't have a Cargo.lock checked in, and even if it did, there are multiple lockfiles I'd like to test with (a "minimum supported versions" one and a "latest versons of all deps" one), so the checked-in version wouldn't be sufficient.

My current non-nix solution involves doing a shallow checkout of every commit and doing a fresh build, which is roughly what `buildRustPackages` would do if I were to naively port my setup into nix. But this is incredibly wasteful, since it rebuilds my entire dep tree on every commit, even though (for normal PRs) the dependency tree doesn't change at all.


kolloch:
Heyo! Couldn't you just do that in the source passed to the tools?

Ericson2314:
Yes, that is better done separately.

apoelstra:
Libraries don't have lockfiles as part of their source. It is a big PITA to "edit the source" of something that is coming from a particular git commit.

Ericson2314:
Are you saying the `tools.nix` stuff will create a lockfile if there is none and you would like to have more control over it? That does help me understand.

apoelstra:
> Are you saying the `tools.nix` stuff will create a lockfile if there is none and you would like to have more control over it? That does help me understand.

I'm saying that I have a set of lockfiles, none of which are from the repo in question, and I want to use those to build a set of source.

PR #278, opened 2023-02-22
Fix git deps with branches vendoring

Ericson2314:
The branch name extraction logic was broken.

As a result of this, we can stop pregenerating the test for this case, demonstrating that `tools.nix` handles it.


Ericson2314:
@balsoft would you mind reviewing this, since I think you contributed large parts of the code I was modifying?

tools.nix
```diff
@@ -232,11 +249,12 @@ rec {
...
           let
             parsed = parseGitSource source;
-            src = builtins.fetchGit {
+            src = builtins.fetchGit ({
               submodules = true;
               inherit (parsed) url rev;
-              ref = attrs.branch or "master";
-            };
+            } // lib.optionalAttrs (parsed ? branch) {
+              ref = parsed.branch;
```
balsoft, line 256: Is the branch always set in the git source URL?

Ericson2314: It isn't, in which case it just takes the default one (remote side `HEAD`). I say this matter for some `main` vs `master` defaults.

balsoft: Yeah, upon a second reading it makes sense.

balsoft (approved):

Ericson2314:
Thanks!

PR #280, opened 2023-03-03
improve git branch parsing and cargo config serialization

apoelstra:
When it comes to git dependencies, cargo is very finicky, and it turns out you basically have to copy the exact logic of 'cargo vendor' to get it to work. This means:

  * The "precise" part of the URL, the part after the #, just gets ignored, for cargo-config purposes.
  * If either "branch" or "ref" are specified as ?key=value parts after the URL, these should be used as the "branch".
  * If neither is specified, the field should be omitted. Similarly, if the URL has no "?rev=" part, this should be omitted (NOT replaced by precise URL part).

I'm unsure what to do if both a tag and a branch are specified, since it looks like cargo vendor treats branch and tag references as disjoint things. Right now I just serialize both fields, but I believe this is incorrect and one should take precedence.

Next, when fetching dependencies via fetchGit, we need to interpret the URL data slightly differently:

  * If a precise part of the URL is given, that should supercede any "?rev=xyz" part of the URL.
  * If a tag or branch is provided we should use it as "ref". Otherwise I just set allRefs = true and hope for the best.

### DIFF
--- a/tools.nix
+++ b/tools.nix
@@ -170,7 +170,9 @@ rec {
         withoutGitPlus = lib.removePrefix "git+" source;
         splitHash = lib.splitString "#" withoutGitPlus;
         preFragment = builtins.elemAt splitHash 0;
-        fragment = builtins.elemAt splitHash 1;
+        fragment = if builtins.length splitHash >= 2
+          then builtins.elemAt splitHash 1
+          else null;
         splitQuestion = lib.splitString "?" preFragment;
         preQueryParams = builtins.elemAt splitQuestion 0;
         queryParamsList = lib.optionals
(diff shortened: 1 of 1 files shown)


apoelstra:
For reference, to justify my claims about what cargo does:

* [here](https://github.com/rust-lang/cargo/blob/16ad1f2945dff5c7f3234257d5ab477ea86b7338/src/cargo/core/source/source_id.rs#L139-L156) is where cargo parses URLs. Later in the same file it returns this data via `fn git_reference`
* ... to [here](https://github.com/rust-lang/cargo/blob/16ad1f2945dff5c7f3234257d5ab477ea86b7338/src/cargo/ops/vendor.rs#L279-L293) which is the cargo-vendor source which produces the cargo config file

apoelstra:
Force-push to change tag refs to be `refs/tags/${parsed.tag}` rather than just `parsed.tag`.

Ericson2314:
Oh thank you very much for this! I think it might fix some issues I too was having.

Ericson2314:
The test failure is just a reformat, I will fix.

PR #281, opened 2023-03-03
allow git deps to have workspaces

apoelstra:
This should be correctly credited to @marius851000 since I copied it directly from https://github.com/kolloch/crate2nix/pull/166/, since it was a narrow part of that changeset that I could understand/justify.

What it does is, in the case that you have a git reference to a crate whose git repo contains workspaces, e.g. [mutagen](https://github.com/llogiq/mutagen/blob/master/Cargo.toml), then rather than trying to directly use the root Cargo.toml (which has no information, just a list of workspaces), it parses it to find the path to the *actual* Cargo.toml for the crate we're trying to depend on, and uses that.

### DIFF
--- a/tools.nix
+++ b/tools.nix
@@ -380,11 +380,28 @@ rec {
                 inherit sha256;
                 inherit (parsed) url rev;
               };
+
+              rootCargo = builtins.fromTOML (builtins.readFile "${src}/Cargo.toml");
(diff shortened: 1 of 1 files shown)


kolloch:
Thank you, @marius851000 and @apoelstra !

Ericson2314:
Could we have a test for this?

Issue #225, opened 2022-05-17
workspace crate source hashes are different between nix 2.3 and later versions


Fuuzetsu:
`crate2nix` generates things like these in `Cargo.nix`:

```nix
      "tsu-mini-std" = rec {
        crateName = "tsu-mini-std";
        version = "0.1.0";
        edition = "2021";
        # We can't filter paths with references in Nix 2.4
        # See https://github.com/NixOS/nix/issues/5410
        src = if (lib.versionOlder builtins.nixVersion "2.4pre20211007")
          then lib.cleanSourceWith { filter = sourceFilter;  src = ../tsu-mini-std; }
          else ../tsu-mini-std;
        dependencies = [
          {
            name = "fxhash";
            packageId = "fxhash";
          }
        ];

      };
```

On `nix` 2.3.x and presumably older, this crate ends up depending on its sources (obviously) like so:

```
[nix-develop]$ nix-store -q -R /nix/store/4y4rb2x1w2ypj5zz1685nhpvhkwwmfhn-rust_tsu-mini-std-0.1.0.drv | tail -n 3
/nix/store/a87gw94pj9l5z1zr4b4bd6hxrcj6y1dg-source
/nix/store/7hvhcn49lynz15s267phv4ph5g2yy7ms-rust_tsu-mini-std-0.1.0.drv
/nix/store/4y4rb2x1w2ypj5zz1685nhpvhkwwmfhn-rust_tsu-mini-std-0.1.0.drv
```

On modern nix (I'm using 2.9 pre), it looks like this:
```
[nix-develop]$ nix-store -q -R /nix/store/68cjkl39fdqf49nmym9d0h1ml44bf9r2-rust_tsu-mini-std-0.1.0.drv | tail -n 3
/nix/store/4i4av9wvdx6djid9phrwqp2vnc2r69pr-tsu-mini-std
/nix/store/giw1vm6k33vv2fc2gg7qlq0sg6wshqx9-rust_tsu-mini-std-0.1.0.drv
/nix/store/68cjkl39fdqf49nmym9d0h1ml44bf9r2-rust_tsu-mini-std-0.1.0.drv
```

[...]

PR #282, opened 2023-03-14
Specify name in cleanSourceWith

bouk:
If you don't specify the name then the src name is used, which will be based on the _unfiltered_ source tree. This means that the output still depends on the input files that are filtered out, and extra builds are done that aren't needed.

I figured this out by debugging why changing any file in my git repo flake caused everything to be rebuilt again, which happens because the src name changes every time _anything_ is changed at all.

Ref: https://nixos.org/manual/nix/stable/language/builtins.html#builtins-filterSource


bouk:
Looking at the implementation of toSourceAttributes I'm not entirely sure now... I might just be running into https://github.com/NixOS/nix/issues/5870

Fuuzetsu:
#225 works around it I think.

PR #285, opened 2023-03-20
Build kinds support

jordanisaacs:
Add build kind support (relying on https://github.com/NixOS/nixpkgs/pull/221440). Uses a new parameter `buildKinds`. Due to dependency derivations, you can only build `["test", "example", "bench"]` together, and `["lib", "bin"]` together. All dependencies will only build `lib`.

A breaking change is the removal of the `runTests` option. Still need to do it, but plan is to move it into a separate tools function that utilizes `buildKinds` under the hood. 

TODO:

1. Move `runTests` into a new tools function
1. Update tests that utilize `runTests` & add new ones for testing benches, examples, etc.
2. Merge new cargo metadata and utilize it in codegen: https://github.com/rust-lang/cargo/issues/11846


jordanisaacs:
Passes all the crate2nix tests!

jordanisaacs:
Been playing around with this in a [project](https://github.com/jordanisaacs/rustix-uring/blob/ecc52382595b73b813336e3a22f0f059a15b8a6d/flake.nix#L126) of mine and it works great. The one issue I see which I am not sure the best way to fix (whether in `buildRustCrate` or `crate2nix`) is the ergonomics of dev dependencies.

Tests/examples/benches that are not the library test, rely on the library itself as a dependency. This is not reflected in the dependency list generated by crate2nix. You can get around this by adding `lib` to the build kinds, BUT then the library gets built with the dev dependencies. Open to suggestions.

crate2nix/src/resolve.rs
```diff
@@ -333,23 +338,33 @@ pub fn double_crate_with_rename() {
...
 }
 
 impl BuildTarget {
     pub fn new(target: &Target, package_path: impl AsRef<Path>) -> Result<BuildTarget, Error> {
         Ok(BuildTarget {
             name: target.name.clone(),
+            kind: target
+                .kind
+                .first()
+                .context("Target shouuld have kind")?
```
flokli, line 359:
```suggestion
                .context("Target should have kind")?
```

crate2nix/templates/Cargo.nix.tera
```diff
@@ -106,6 +106,11 @@ rec {
           {
             name = {{ bin.name }};
             path = {{ bin.src_path }};
+            kind = {{ bin.kind }};
+            test = {{ bin.test }};
+            # TODO (when cargo merges metadata):
+            # harness
+            # bench
```
flokli, line 113: Your PR https://github.com/rust-lang/cargo/pull/11847 got closed due to inactivity. Do you still intend to resume this?

nix/sources.json
```diff
@@ -48,15 +48,15 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "nixpkgs": {
```
flokli, line 50: :eyes:

flokli:
@jordanisaacs did you keep tinkering with this? We'd like to build benchmarks with crate2nix for [Tvix](https://tvix.dev/), so I'm curious to see if I can help out somehow.

jordanisaacs:
@flokli It was a pretty big change that needs changes in three projects (crate2nix, cargo, and nixpkgs) and it didn't get any reviews so stopped pushing it forward. I would love to see this stuff get merged though.

I see the way forward being getting the nixpkgs PR merged first. That needs reviewers of course. Then we can revisit crate2nix and what the needs merged into cargo (what metadata do we need).

jordanisaacs:
@flokli It was a pretty big change that needs changes in three projects (crate2nix, cargo, and nixpkgs) and it didn't get any reviews so stopped pushing it forward. I would love to see this stuff get merged though.

I see the way forward being getting the nixpkgs PR merged first. That needs reviewers which you could help with. Then we can revisit crate2nix and what we need merged into cargo (e.g. what metadata do we need).

jordanisaacs:
I do need to rebase the nixpkgs commit though. Will try and do it sometime this week.

PR #288, opened 2023-08-26
src/render.rs: handle CfgExpr::Any(vec![])

amjoseph-nixpkgs:
Apparently `CfgExpr::Any(vec![])` occurs in the wild.  Encountered while crate2nixifying `cosmic-comp`.


Ericson2314:
Thanks @amjoseph-nixpkgs. Would you mind creating a test for this? And could you do `All` with length 0 too?

Ericson2314:
Oh #289 right after this included `All` and the tests, by shear coincidence. Going with that instead of this.

PR #295, opened 2023-09-15
Add flake.nix

bouk:
This adds a flake.nix with an overlay and a package. The nixpkgs is pinned to the one used by the rest of the repository, so its nixpkgs version is not managed separately.

After this PR `crate2nix` can be run by calling `nix run github:nix-community/crate2nix`. Crate2nix can also be used via the overlay by using `crate2nix.overlays.default`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,44 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1694529238,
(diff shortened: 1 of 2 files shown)


kolloch:
Hi @bouk, thank you!

could you please reformat it with https://github.com/nix-community/crate2nix/blob/master/nixpkgs-fmt.sh?

while it is a quite minimal wrapper around existing config, it would be nice if this was checked in CI so that we don’t break it accidentally.

tilpner:
It would be nice if this could export `tools.nix` in some way. Not sure where it fits in exactly, maybe that can wait until after flake-schemas is a thing, and you can make up attribute semantics.

kolloch:
Thanks!

PR #298, opened 2023-09-20
Git workspaces

tilpner:
This adds support for using workspaces in git dependencies with inherited fields, building upon #166 and #281.
An example project which doesn't build on master, but works with this PR, can be found [here](https://github.com/tilpner/crate2nix-testcases/tree/9fb2f38e6dccc520bfec80a8d7b42fabf05de67d/git-workspaces).

#166 takes care of exposing workspace members at the top level of a vendoring entry, so that cargo can find it, but it abducts the crate from its workspace without adjusting the manifest.
For example, the [sqlx-postgres manifest](https://github.com/launchbadge/sqlx/blob/35ba8e2b93f085bbd3b40d31c2f7b01a9911a9de/sqlx-postgres/Cargo.toml#L38-L46) inherits dependencies and metadata (edition, license) from the workspace manifest. These fields can't be filled in by cargo after copying the member crate out of the workspace, so this PR fills in that information before/while copying it out.

It does add a dependency on cargo's unstable API, which I found preferable to reimplementing the manifest inheritance on top of `cargo_metadata`, but I'd understand if it's an unacceptable choice for crate2nix. cargo (in the most recent version, which might not be required) also expects a very recent rustc.

The libgit2-sys dependencies are currently duplicated, due to lack of a central place to put them. If tools.nix could import default.nix, this wouldn't be a problem, but that seems to have been avoided so far, and would probably cause changes to the public interface.

If the commit message convention is out of place, or the drive-by "improvements" (don't shadow nix, use stable CLI) don't fit into this PR, we can change/drop them.

### DIFF
--- a/crate2nix/Cargo.lock
+++ b/crate2nix/Cargo.lock
@@ -3,6 +3,24 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if",
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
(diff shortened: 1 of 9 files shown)


kolloch:
Heyo @tilpner, I am not sure when I have time to think about this. But I already merged some of your commits on master, so thanks a lot!

hallettj:
Hi! I really appreciate the work done in this PR! After the changes that have already been merged I still have a desire for an option to get dependencies from workspaces via git in IFD mode. So I've merged the PR branch with the latest crate2nix code, and fixed things up. I think it's in a good state to review:

https://github.com/hallettj/crate2nix/tree/git-workspaces

That branch depends on #382 which should be merged first.

hallettj:
I've opened a new PR with my updated fixes for git workspaces in #393

PR #301, opened 2023-10-10
Run tests in debug mode

edef1c:
Previously, crate2nix tests were run equivalently to `cargo test --release`, which is a somewhat unusual default configuration.

It might be worth having a flag to tweak this behaviour, similar to Cargo. A more ambitious version yet would permit running both debug and release versions of the tests. I'm not sure what the best interface for either of those is.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -2444,6 +2444,7 @@ rec {
             (
               _: {
                 buildTests = true;
+                release = false;
               }
             );
           # If the user hasn't set any pre/post commands, we don't want to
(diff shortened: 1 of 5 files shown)


kolloch:
Hi @edef1c!

I also think it should be configurable. Usually, I execute tests locally with cargo in debug mode but then in CI in release mode. Depends quite on the test. Obviously, model tests or the like profit from release mode more than others.

I haven't worked on crate2nix for a while and will first invest in flakifying it and making it work on Mac OS again (which I am using currently). But I'd be interested in hearing how such an interface could look like!

PR #317, opened 2024-01-03
Use flake-parts and flake-compat

kolloch:
...for default.nix and shell.nix.

Ultimately, I want to get rid of sources.nix/niv and make the flake inputs the source of truth.


Ericson2314:
I would not like to get rid of the old `default.nix`. The current issues around especially platform support with Flakes (open world vs closed world) make it a clear loss in some circumstances.

I would rather have an optional flake that wraps the `default.nix`. The `default.nix` can no longer have Niv-supplied default parameters (just use the Flake if you want that) so we can still get rid of Niv if you like. 

(See how we recently split out https://github.com/NixOS/nix/blob/master/package.nix for a similar, Flake-on-the-outside, old-school-on-the-inside approach.)

kolloch:
Hey @Ericson2314,

I merged this for now before seeing your comment.

Basically, I want to push this to a more modern approach. Can you elaborate which `default.nix` you deem important? I guess stuff like overriding system?

I’d like to use as much from the modern flake infrastructure as possible - without unnecessarily breaking valid use cases, of course.

kolloch:
Looking at your example, you are NOT opposed to using flake-compat for default.nix. But would like to have an inner “traditional” layer mostly building crate2nix itself?

I wouldn’t have big concerns for not using flake compat for default.nix, though. I’d keep it for shell.nix.

Ericson2314:
Flake compat does not address my concerns, because it doesn't help with the finitely enumerated system problem. The problem is the structure of the flake nix expression, not the fact that it is a "flake". Flake compat cannot fix that.

kolloch:
Can you speak in less abstract terms that I might understand?

Doesn't that basically mean "yes" to the question:

> Basically, I want to push this to a more modern approach. Can you elaborate which default.nix you deem important? I guess stuff like overriding system?

I would like to understand your use cases.

kolloch:
This might make you happy: https://github.com/nix-community/crate2nix/pull/320

Ericson2314:
(It did make me very happy, thanks!)

PR #320, opened 2024-01-04
Provide callPackageable default.nix again

kolloch:
After concerns raised here: https://github.com/nix-community/crate2nix/pull/317#issuecomment-1876041332


Ericson2314:
Thank you so much @kolloch! This looks perfect, exactly what I wanted.

PR #325, opened 2024-01-08
Fix systems

JanotLeLapin:
I noticed the `systems` variable contained `"x86_64-linux"` twice, and I believe this might be an error.


kolloch:
Thank you!

PR #341, opened 2024-04-03
Stop fabricating PackageId and look it up instead

### LINKED ISSUE #340: Crate hash loading from Cargo.lock is broken in Rust 1.77
https://github.com/rust-lang/cargo/pull/12914 changed how package ID looks like, now it's something like this:

```diff
-"bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)"
+"https://github.com/rust-lang/crates.io-index#bitflags@1.0.4"
```

This changed how `PackageId` is represented:
https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.PackageId.html

> It is possible to inspect the repr field, if the need arises, but its precise format is an implementation detail and is subject to change.

At the same time `lock.rs` creates them from parts: https://github.com/nix-community/crate2nix/blob/e494b56b7445ba59ef049bcfdc889f5c558768d4/crate2nix/src/lock.rs#L42

As a result mapping between "shortened" and "full" ids breaks, and it fails to find any hashes. It's not the end of the world - `crate2nix` can fetch crates on it's own and place them in a file, but this is very slow if you have a lot of them.

pacak:
- lock.rs: get rid of confusing Metadata type alias
- IndexedMetadata::new_from_merged doesn't consume MergedMetadata
- Lookup PackageId in resolved metadata

Fixes #340

### DIFF
--- a/crate2nix/src/lock.rs
+++ b/crate2nix/src/lock.rs
@@ -8,6 +8,8 @@ use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
 
+use crate::metadata::MergedMetadata;
+
 impl EncodableResolve {
     pub fn load_lock_file(path: &Path) -> Result<EncodableResolve, Error> {
         let config = &std::fs::read_to_string(path)
@@ -27,8 +29,19 @@ impl EncodableResolve {
 
     pub fn get_hashes_by_package_id(
         &self,
+        metadata: &MergedMetadata,
         hashes: &mut HashMap<PackageId, String>,
     ) -> Result<(), Error> {
+        let mut package_id_by_source = HashMap::new();
+        for p in &metadata.packages {
+            let Some(ref source) = p.source else {
+                // local crate
+                continue;
+            };
+            let key = (p.name.as_str(), source.repr.as_str(), p.version.to_string());
+            package_id_by_source.insert(key, &p.id);
+        }
+
         for EncodableDependency {
             name,
             version,
@@ -37,13 +50,17 @@ impl EncodableResolve {
             ..
         } in self.package.iter()
         {
-            if let (Some(source), Some(checksum)) = (source, checksum) {
-                let package_id = PackageId {
-                    repr: format!("{} {} ({})", name, version, source),
-                };
-                if checksum != "<none>" {
-                    hashes.insert(package_id, checksum.clone());
-                }
+            let Some((source, checksum)) = Option::zip(source.as_ref(), checksum.as_ref()) else {
+                continue;
+            };
+            if checksum == "<none>" {
+                continue;
+            }
+
+            if let Some(package_id) =
+                package_id_by_source.get(&(name.as_str(), source.as_str(), version.clone()))
+            {
+                hashes.insert((*package_id).clone(), checksum.clone());
             }
         }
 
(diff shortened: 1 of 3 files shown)


kolloch:
Hey @pacak,

thanks for trying to address this! I believe the general direction is the right one, even though I forgot much about the context in the code since it is quite a long time that I wrote (parts of) it. Let me comment on the easy to agree with parts:

> * lock.rs: get rid of confusing Metadata type alias

Note that this is code copied from elsewhere, so potentially it would be worth it to keep it closer to the source, but hey, I merged it.

> * IndexedMetadata::new_from_merged doesn't consume MergedMetadata

Makes sense, since it doesn't reuse anything anyways

kolloch:
I noticed that you removed quite some tests, can you comment on that?

crate2nix/src/lock.rs
```diff
@@ -37,13 +50,16 @@ impl EncodableResolve {
...
         } in self.package.iter()
         {
-            if let (Some(source), Some(checksum)) = (source, checksum) {
-                let package_id = PackageId {
-                    repr: format!("{} {} ({})", name, version, source),
-                };
-                if checksum != "<none>" {
-                    hashes.insert(package_id, checksum.clone());
-                }
+            let Some((source, checksum)) = Option::zip(source.as_ref(), checksum.as_ref()) else {
```
kolloch, line 53: Nice! Didn't know that one.

crate2nix/src/lock.rs
```diff
@@ -66,88 +82,6 @@ impl EncodableResolve {
     }
 }
 
-#[test]
-fn test_no_legacy_checksums() {
```
kolloch, line 70: Please don't delete them, rather add some for the new cases :)

crate2nix/src/lock.rs
```diff
@@ -27,8 +29,19 @@ impl EncodableResolve {
 
     pub fn get_hashes_by_package_id(
         &self,
+        metadata: &MergedMetadata,
         hashes: &mut HashMap<PackageId, String>,
     ) -> Result<(), Error> {
+        let mut cache = HashMap::new();
```
kolloch, line 35: I would prefer some thing like `packageIdBySource`

pacak:
> Note that this is code copied from elsewhere, so potentially it would be worth it to keep it closer to the source

I don't have strong opinions about it, but knowing that we are working with `cargo-metadata`- I assumed that it was `Metadata` from that crate and tried to use it accordingly, wasting some time. Just wanted to save this trouble for other people.

> I noticed that you removed quite some tests, can you comment on that?

New code relies on info from `cargo-metadata`, tests don't have it included. Faking might work but it isn't the best idea. For as long as this crate uses itself to generate its own `Cargo.nix` this serves as a test for current lock file format. Legacy might or might not work, but using the most recent `crate2nix` with very old rust tool chain...

kolloch:
I would like the code supporting for the old legacy formats tested.

Not sure how to decide that it is not needed anymore, then the code could be removed and the tests as well. But I'd not like to do that without a deprecation period.

pacak:
> Not sure how to decide that it is not needed anymore, then the code could be removed and the tests as well. But I'd not like to do that without a deprecation period.

Comment calling it legacy was added in January 2020, so at most 1.42 - 35 rust releases ago. `crate2nix` requires at least 1.56 to compile - it uses 2021 edition, actual MSRV might be higher due to dependencies. I'd say it is safe to drop the support right now but it goes outside of the scope for this pull request.

kolloch:
Still, can we add tests for the new stuff?

Sure, it will probably fail elsewhere if there is a problem.

But it is usually easier to understand with more specific tests.

pacak:
> Still, can we add tests for the new stuff?

Will look into that when I have free time.

flokli:
@kolloch can you publish a release with this?

kolloch:
Done, hopefully not too rushed.

flokli:
Thanks! I opened https://github.com/NixOS/nixpkgs/pull/303309 so nixpkgs has the fix too.

Ten0:
Not sure if that's relevant but IFD code seems to still generate such IDs with old format to store them in `crate-hashes.json`, while Rust code now generates those with the new format.
Looking at https://github.com/nix-community/crate2nix/issues/340#issue-2206506647, it seems that it may be what's causing https://github.com/nix-community/crate2nix/issues/348.

pacak:
I'm seeing stuff like this which suggests new format. Old format comes from old version of cargo.
```
  "git+https://github.com/nix-rust/nix?rev=fbdac70c78975fe7c75b882337bf9f40a639f51a#0.25.0": "1jvls4d90ns5qwqab1a2hcyff2dlls1q4i5xl6cs09a371g5h7b3",
```

pacak:
And in case of #340 it was working, but wasn't able to use cache so I don't think it's related.

pacak:
> Isn't it generated on the nix side in the case of IFD?

https://github.com/nix-community/crate2nix/blob/cf034861fdc4e091fc7c5f01d6c022dc46686cf1/tools.nix#L246-L249

Looks like it indeed. That's not going to work...

flokli:
I just ran `crate2nix generate --all-features` with the more recent rust version (not inside IFD), and `crate2nix` started to manually add *all* crate hashes into `crate-hashes.json`, not just for the ones for which there are no hashes in `Cargo.lock` - so definitely something still seems to be off.

Note it however still builds for me…

pacak:
> I just ran crate2nix generate --all-features with the more recent rust version (not inside IFD)

Is this using `crate2nix` after the change, 0.14.0? Anything public or any reproduction? Works for me...

flokli:
Hmmh, no, indeed I'm still on 0.13.0 for some reason. I'll get this bumped and will report back.

flokli:
Indeed, this fixed it. Sorry for the noise!

Ten0:
> Looks like it indeed. That's not going to work...

I can still see [this on the current master](https://github.com/nix-community/crate2nix/blob/8537c2d7cb623679aaeff62c4c4c43a91566ab09/tools.nix#L247-L248). Is there no plan to fix this? If not, should that be [documented](https://nix-community.github.io/crate2nix/00_guides/31_auto_generating/)?

pacak:
> Is there no plan to fix this? If not, should

Can you make a small example that would reproduce the problem?

pacak:
> Should I open a separate issue?

Sure. Would be great if it had an example you could just copy-paste or clone and see it panic.

PR #357, opened 2024-08-03
Fix: Use mkDerivation with src instead of runCommand for test derivation

Kranzes:
The problem with using runCommand and recreating the src directory with
lndir is that it changes the file types of individual files, they will
now be a symlink instead of a regular file. If you have a crate that tests
that a file is of regular type then it will fail inside the crate2nix derivation.

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -3248,52 +3248,41 @@ rec {
               testPostRun
             ]);
         in
-        pkgs.runCommand "run-tests-${testCrate.name}"
-          {
-            inherit testCrateFlags;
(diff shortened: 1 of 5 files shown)


apoelstra:
As a user with a ton of disk space I would prefer this (many of my uses of crate2nix involve nasty hacks to un-symlink files so that I can either modify them or just pass "file exists and is `-f` checks".

But this will use more disk space and slow things down, possibly by a lot on large repos.

Kranzes:
I think people should filter their src's properly with something like lib.fileset. They shouldn't be copying over the entire repo to the nix store.  It's not that bad IMO.

PR #360, opened 2024-08-26
Document using `rust-overlay`

drakon64:
Add some documentation for using `rust-overlay` with IFD and Flakes. This is what worked for my setup, adjusted to use `flake-parts` instead of `flake-utils`.

Fixes https://github.com/nix-community/crate2nix/issues/359.


kolloch:
Thank you so much!

PR #363, opened 2024-09-28
Add docs about custom toolchains

surma:
I have found no documentation on how to use `crate2nix` with a custom toolchain.

After a bunch of fiddling I came up with *an* solution and thought I’d provide a PR to add this as documentation.

That being said: I am not sure this is the _intended_ way to provide a custom toolchain. I am happy to adapt the PR if anyone is so kind to show me a better way to do this.


kolloch:
Thanks! I just merged another approach as well :)

PR #366, opened 2024-10-24
Adding support for private registries

P-E-Meunier:
Private registries were currently handled by generating code that was referencing the cache in ~/.cargo. This commit adds support to generate correct code for them.


kolloch:
The tests seem to fail…

P-E-Meunier:
Indeed, because they're comparing the produced file with a file that doesn't handle private registries. Where do I fix them?

kolloch:
Did you check this out?

https://nix-community.github.io/crate2nix/50_contributing/05_code/

PR #368, opened 2024-10-24
Adding the dependency to libsecret on Linux (for private registries) 

P-E-Meunier:
On Linux, `cargo-metadata` has a hidden dependency on Libsecret via `dlopen`. This PR sets the path correctly in the ELF.


kolloch:
Thank you :)

PR #372, opened 2024-12-04
Fix build-dependencies resolution when cross-compiling

pnmadelaine:
This fixes some errors in the resolution of build-dependencies when cross-compiling. All dependencies were computed using the host platform, even build-dependencies which are then compiled for the build platform. I added a minimized example that highlights the issue.


Ericson2314 (commented):
Oh this looks very nice!

Ericson2314:
@alyssais you have any thoughts on the `name` field? I forgot what Nixpkgs is doing atm, but I do think we want to keep the target stuff in sync.

alyssais:
Using rustcTarget as this does seems like the right thing.

PR #374, opened 2024-12-17
Make all functions overridable

P-E-Meunier:
Currently, most functions take `pkgs` as an argument, and fail to use some of their other arguments, effectively replacing them with `pkgs.[blah]`. This commit fixes that.


Ericson2314:
Makes sense to me. Should we use `_pkgs` or something as the new name to indicate that it should *only* be used to define default arguments?

PR #375, opened 2024-12-18
Fix cargo tests & clippy warnings

pnmadelaine:


pnmadelaine:
@Ericson2314 well it's not 😁 the cargo tests were broken, but CI does not run them!

Ericson2314:
Oh! Lovely....

Well, thank you then!

PR #381, opened 2025-01-23
fix parsing checksums from v1 manifests with latest cargo

hallettj:
Crate2nix identifies packages using package ID strings reported by the `cargo metadata` subcommand. Cargo describes these strings as internal identifiers subject to change. And in fact the string format did change sometime between cargo 1.76 and cargo 1.78.

In cargo 1.76 package ID strings look like this:

```
aho-corasick 1.1.3 (registry+https://github.com/rust-lang/crates.io-index)
```

In later cargo versions the format changed to look like this:

```
registry+https://github.com/rust-lang/crates.io-index#aho-corasick@1.1.3
```

The change is fine as long as package IDs are used as opaque identifiers in memory. Unfortunately one use of package IDs is to look up checksums in the legacy v1 `Cargo.lock` manifest format. That format stores checksums in a map at the end of the file that looks like this:

```toml
[metadata]
"checksum ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)" = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
"checksum atty 0.2.13 (registry+https://github.com/rust-lang/crates.io-index)" = "1803c647a3ec87095e7ae7acfca019e98de5ec9a7d01343f611cf3152ed71a90"
"checksum autocfg 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)" = "1d49d90015b3c36167a20fe2810c5cd875ad504b39cff3d4eae7977e6b7c1cb2"
```

Being a legacy format the keys in these checksum maps will always use the old package ID format. The problem becomes quickly apparent because nix-test-runner - which is used in the crate2nix test suite - happens to use a v1 manifest. That leads to tests and the devShell breaking when attempting to update crate2nix flake inputs.

Currently crate2nix assumes that package IDs read from `cargo metadata` output can be used as lookup keys in the v1 checksum map. But when a recent version of cargo is installed that assumption is broken.

This PR fixes the problem by constructing checksum lookup keys explicitly. This should make checksum lookups work regardless of which cargo version is installed.


hallettj:
Tests are not currently passing for this PR. I have put test fixes in #380 - I didn't want to duplicate those changes here.

hallettj:
To produce this fix I referenced the corresponding implementation from rustsec/cargo-lock:

https://github.com/rustsec/rustsec/blob/bd6fb0fba41246ed860e0e8374d8f31aceabc8f9/cargo-lock/src/metadata.rs#L24

Issue #359, opened 2024-08-22
Build fails when using `rust-overlay` due to some dependencies being built with Nixpkgs Rust when cross-compiling


drakon64:
I have a simple derivation using `crate2nix`:

```nix
{
  pkgs,
  crate2nix,
  system,
  src,
  ...
}:
let
  rustToolchain = pkgs.rust-bin.stable."1.80.1".default;

  buildRustCrateForPkgs =
    pkgs:
    pkgs.buildRustCrate.override {
      rustc = rustToolchain;
      cargo = rustToolchain;
    };

  generatedCargoNix = crate2nix.tools.${system}.generatedCargoNix {
    name = "package";
    inherit src;
  };

  cargoNix = import generatedCargoNix {
    inherit pkgs buildRustCrateForPkgs;
  };
in
cargoNix.rootCrate.build
```

When I try to build this, I get this error:

```
error[E0514]: found crate `derive_builder_macro` compiled by an incompatible version of rustc
   --> src/lib.rs:718:1
    |
718 | extern crate derive_builder_macro;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the following crate versions were found:
            crate `derive_builder_macro` compiled by rustc 1.80.1 (3f5fd8dd4 2024-08-06) (built from a source tarball): /nix/store/ay24n13hpxymp9wvrw08a0qh5fwzkhlx-rust_derive_bui>
    = help: please recompile that crate using this compiler (rustc 1.80.1 (3f5fd8dd4 2024-08-06)) (consider running `cargo clean` first)

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0514`.
```

If I downgrade the Rust overlay to 1.80.0, I get the same error but with 'this' compiler correctly appearing as 1.80.0.

It seems like `crate2nix` is still building some of the dependencies with the Nixpkgs version of Rust.

In my Flake, `pkgs` is defined as follows:
```nix
pkgs = import nixpkgs {
  inherit system;

  overlays = [ (import rust-overlay) ];

  crossSystem = {
    config = "x86_64-unknown-linux-musl";
  };
};
```

drakon64:
I cannot reproduce this when commenting out `crossSystem.config`, or changing it to `x86_64-unknown-linux-gnu`

drakon64:
[...]

PR #384, opened 2025-02-09
docs(guide): :memo: fix `rust-overlay` guide

BeiyanYunyi:
The guide of using `rust-overlay` was committed at #360. However, it has a typo and thus cannot build the derivation successfully, which confused me for a while.

After searching for examples, I found in #359, author used `generatedCargoNix` instead of `appliedCargoNix`, then I switched to it, make my crate successful to build.


kolloch:
Thank you :)

PR #392, opened 2025-03-08
expand wildcards in workspace member paths

hallettj:
This fixes a problem that comes up when referencing a dependency via git where the dependency is in a workspace, and the workspace manifest lists members using wildcards / file globs. An example is [ndc-sdk-rs](https://github.com/hasura/ndc-sdk-rs/blob/main/Cargo.toml) whose `Cargo.toml` defines a virtual workspace like this:

```toml
[workspace]
resolver = "2"
members = ["crates/*"]
```

When using crate2nix in IFD mode referencing dependencies in this workspace fails because `tools.nix` expects workspace members to be given as exact paths, without wildcards. You get an error of this form:

```
No such file or directory, crates/*
```

This PR adds a function to produce a list of exact paths given a path that might contain wildcards, and a directory to search in. It modifies the logic that looks up crates in a workspace in a git dependency to use the new function.

I included a number of unit tests to try to verify the functionality of glob expansion. I think an integration test would be a good idea, but that requires a dependency that can be referenced via git. The best example I know of is ndc-sdk-rs which I mentioned above. It's non-trivial so I was hesitant to add the necessary build time to the test suite in crate2nix.

### DIFF
--- a/crate2nix/templates/nix/crate2nix/tests/default.nix
+++ b/crate2nix/templates/nix/crate2nix/tests/default.nix
@@ -8,6 +8,7 @@ let
     "enableFeatures"
     "expandFeatures"
     "packageFeatures"
+    "pathsFromPathPattern"
   ];
   testsInFile = f:
     let
--- a/crate2nix/templates/nix/crate2nix/tests/pathsFromPathPattern.nix
+++ b/crate2nix/templates/nix/crate2nix/tests/pathsFromPathPattern.nix
@@ -1,0 +1,55 @@
+{ lib, ... }:
+let
+  module = import ../../../../../nix/lib/paths-from-path-pattern.nix { inherit lib; };
+  pathToRegex = module.pathToRegex;
+  pathsFromPathPattern = module.pathsFromPathPattern;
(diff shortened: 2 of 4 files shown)


Pacman99:
I just opened a PR that I think has a simpler solution for this that also solves another problem that I describe in that PR: https://github.com/nix-community/crate2nix/pull/394

hallettj:
> I just opened a PR that I think has a simpler solution for this that also solves another problem that I describe in that PR: #394

Oh yes, I see how avoiding implementing file path wildcards in Nix is simpler! I notice that #394 ignores the `workspace.members` property of the workspace manifest, and instead gathers all `Cargo.toml` files in the source tree. My first thought was that might cause problems with projects with example or test crates that aren't intended to be distributed. But then I saw this comment which seems to indicate that maybe #394 is a better match to Cargo's behavior than either this PR, or the preexisting behavior: https://github.com/rust-lang/cargo/issues/1462#issuecomment-87002475

hallettj:
Closing in favor of #394

PR #400, opened 2025-08-06
Fix IFD git dependency package ID format discrepancy with recent cargo versions

landaudiogo:
As described in [PR](https://github.com/nix-community/crate2nix/pull/393#issuecomment-2746337794) and [PR](https://github.com/nix-community/crate2nix/pull/381#issue-2806252718), crate2nix assumes a package ID format that is only compatible with older versions of cargo. 

This PR changes `tools.nix` to write the package ids to `crate-hashes.json` in the same format as the version of cargo available to crate2nix fixing the observed discrepancy. I have created a [repository](https://github.com/landaudiogo/crate2nix-IFD-issue?tab=readme-ov-file) which motivates the problem, and shows how the proposed fix solves it.

### DIFF
--- a/tools.nix
+++ b/tools.nix
@@ -61,7 +61,7 @@ rec {
         cp ${vendor.cargoConfig} $out/cargo/config
 
         crate_hashes="$out/crate-hashes.json"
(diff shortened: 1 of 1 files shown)


landaudiogo:
Thanks! Just added the commit that formats the PR

Issue #402, opened 2025-08-25
makeDefaultTarget should not hard-code env


sepointon:
# Demonstration

Repository: https://gitlab.com/sampointon/crate2nix-cross-missing-target

Run `nix build -f default.nix`.

Build log showing the failure:

```
Running phase: unpackPhase
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking source archive /nix/store/xlsjn3anddd5jxzjs2bk3ifgbckqj3az-getrandom-0.3.3.tar.gz
source root is getrandom-0.3.3
setting SOURCE_DATE_EPOCH to timestamp 1153704088 of file "getrandom-0.3.3/tests/mod.rs"
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running cd .
Building build.rs (getrandom)
Running rustc --crate-name build_script_build build.rs --crate-type bin -C opt-level=3 -C codegen-units=1 --edition 2021 --cfg feature="default" --out-dir target/build/getrandom --emit=dep-info,link -L dependency=target/buildDeps --cap-lints allow --color always
cargo:rerun-if-changed=build.rs
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
Building src/lib.rs (getrandom)
Running rustc --crate-name getrandom src/lib.rs --out-dir target/lib -L dependency=target/deps --cap-lints allow -C opt-level=3 -C codegen-units=1 --remap-path-prefix=/build=/ --extern cfg_if=/nix/store/d8va8c8kwbz66ppz5h0j4xabz0x48233-rust_cfg-if-1.0.3-wasm32-wasi-lib/lib/libcfg_if-a0f22aa586.rlib --cfg feature="default" --target wasm32-wasip2 --edition 2021 -C metadata=d5632c2b3e -C extra-filename=-d5632c2b3e --crate-type lib -L /build/getrandom-0.3.3/target/build/getrandom.out --color always
[...]

Issue #370, opened 2024-11-05
Some crates.io hashes cannot be found in Cargo.lock


YPares:
I'm running into this:

```
Running phase: unpackPhase
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking source archive /nix/store/qc74kg08frlgdjb5fjmwgvb0snpf45lx-unarchiver
source root is unarchiver
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
++ crate2nix generate -f ./Cargo.toml -o Cargo-generated.nix -h /nix/store/2mwh4sn153gar2r2ay6x43l7752pr3v3-rustnix-crate2nix/crate-hashes.json
Did not find all crates.io hashes in Cargo.lock. Hashes for e.g. addr2line 0.24.2, adler 1.0.2, adler2 2.0.0, aho-corasick 1.1.3, android-tzdata 0.1.1, android_system_properties 0.1.5, async-channel 2.3.1, async-stream 0.3.6, async-stream-impl 0.3.6, async-trait 0.1.83 are missing.
This is probably a bug.
Prefetching    1/379: https://static.crates.io/crates/encoding_rs/encoding_rs-0.8.35.crate
Error: while prefetching crates for calculating sha256: While spawning 'nix-prefetch-url https://static.crates.io/crates/encoding_rs/encoding_rs-0.8.35.crate --name encoding_rs-0.8.35': No such file or directory (os error 2)
crate2nix failed.
```

It's difficult to make a minimal reproduction but just in case is it a known error?

YPares:
@danieldk Thanks! I'll try that. I suppose then that I should re-run `crate2nix` manually everytime a git-provided crate has to change?
I'm not sure it's part of crate2nix documentation, could it be added? I think notable here: https://nix-community.github.io/crate2nix/00_guides/31_auto_generating/

PR #403, opened 2025-08-26
Use rust.platform.env, where set, in conditional compilation

sepointon:
Commits have useful messages. The first one only adds tests, one of which is expected to fail due to the bad current behaviour, which is then fixed by the second. #402 is the motivator and has context.

This is _not_ mergeable as is, so it's marked as a draft: the pinned Nixpkgs is old and doesn't deal with Wasi libraries, causing the tests to fail. I have not succeeded in bumping the pin due to running in to #370. Relatedly, I wasn't able to check that the tests fail without the second, fixing commit, due to it falling over for other reasons first. But I've verified that it fixes the problem I was running in to in real code.


sepointon:
https://github.com/NixOS/nixpkgs/pull/531986 has given Nixpkgs knowledge of `target_env`, so this is almost good to go. I'm just running the tests locally and then I'll push and mask as ready.

One minor question is going to be how long to keep the fallback for a missing `platform.env` (that is, when do we assume that all users have updated their Nixpkgs to include that commit?), but the fallback branch oughtn't do any harm and so that can keep until someone gets sufficiently annoyed to clean it up.

sepointon:
Well, the tests failed for some inscrutable flakes-related reason. Let's see if CI likes it any better.

sepointon:
I think the only thing blocking merging now is the Nixpkgs commit making it into the channel crate2nix is tracking, and then bumping the lock.

Issue #287, opened 2023-05-04
feature request: be able to override sourceFilter


bbigras:
It would be nice to be able to use our own sourceFilter.

PR #405, opened 2025-09-18
Allow customizing sourceFilter to build file

giodamelio:
Allow passing in a `sourceFilter` argument into the `Cargo.nix`.

See #287 

cc @flokli

### DIFF
--- a/crate2nix/Cargo.nix
+++ b/crate2nix/Cargo.nix
@@ -26,6 +26,8 @@
   ? if builtins.pathExists ./crate-config.nix
     then pkgs.callPackage ./crate-config.nix {}
     else {}
+  # Custom source filter function
+, sourceFilter ? null
 }:
 
 rec {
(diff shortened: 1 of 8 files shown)


giodamelio:
I need to test this out a little better, but no tests are failing and it seems to work with a quick test.

flokli:
Thanks! Best add a test, using the filter.

See https://github.com/nix-community/crate2nix/blob/be31feae9a82c225c0fd1bdf978565dc452a483a/tests.nix, you should be able to use customBuild to call into Cargo.nix with the filter.

As for checking whether the filter is applied, maybe use a build.rs script checking whether a file exists in the sources, and based on that have the main binary have a different output?

aanderse:
thanks for implementing this @giodamelio! very useful :tada:

is this something you intend to continue working on?

emanueljg:
In lieu of this PR being merged, to people encountering the same annoying behavior that I did of unneccesary derivation rebuilds, you can override the root crate src with your own and tighten the derivation, e.g. using `lib.fileset`.

```nix
            cargoNix.rootCrate.build.override {
              runTests = true;
              runClippy = true;
              features = [
                # ...
              ];
              crateOverrides = pkgs.defaultCrateOverrides // {
                YOUR_ROOT_CRATE = prev: {
                  # ...
                  src =
                    let
                      fs = lib.fileset;
                    in
                    fs.toSource {
                      root = ./.;
                      fileset = fs.unions [
                        ./src
                      ];

                    };
```

giodamelio:
@aanderse Unfortunately I don't have the time right now to get this over the finish line. I would be happy for someone else to continue my work though.

@emanueljg That is a great workaround. Much better then what I currently do which is just manually editing the crate.nix each time I change my deps 😂

PR #430, opened 2026-01-27
fix: resolve workspace inheritance for git dependencies

domenkozar:
## Summary

- Reconstruct standalone `Cargo.toml` from `cargo_metadata::Package` at generate time for git deps that use `workspace = true`, embedding it in the generated `Cargo.nix` as `resolvedCargoToml`
- Patch the correct subdirectory `Cargo.toml` (not the repo root) via `resolvedCargoTomlSubDir`
- Add Nix-side workspace inheritance resolution in `tools.nix` for the vendoring/sandbox path (`generatedCargoNix`), with `hasWorkspaceTrue`, `resolveDependency`, and `resolveWorkspaceInheritance` helpers
- Add sample project `bin_with_workspace_git_dep` using `nix-conf-parser` from `cachix/devenv` (which uses workspace inheritance for package fields and dependencies)
- Add 20 Nix unit tests (`tools-tests.nix`) and 9 Rust unit tests for the new functionality

## Test plan

- [x] `cargo test` — 23 Rust tests pass
- [x] `./nix-test.sh ./tools-tests.nix` — 20 Nix unit tests pass
- [x] `./nix-test.sh ./crate2nix/templates/nix/crate2nix/tests/default.nix` — 38 template tests pass
- [x] `nix build -f ./tests.nix bin_with_workspace_git_dep` — end-to-end sample project builds and runs
- [x] `./regenerate_cargo_nix.sh` — all Cargo.nix files regenerate cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crate2nix/templates/Cargo.nix.tera
```diff
@@ -168,6 +168,12 @@ rec {
           sha256 = {{ crate.source.Git.sha256 }};
           {%- endif %}
         };
+        {%- if crate.source.Git.resolved_cargo_toml %}
+        resolvedCargoToml = {{crate.source.Git.resolved_cargo_toml}};
```
Mic92, line 172: is this needed for the json version as well?

crate2nix/src/resolve.rs
```diff
@@ -511,11 +517,27 @@ impl ResolvedSource {
         };
         url.set_query(None);
         url.set_fragment(None);
+
+        let (resolved_cargo_toml, resolved_cargo_toml_sub_dir) = {
+            let manifest = package.manifest_path.as_std_path();
+            let content = std::fs::read_to_string(manifest).ok();
+            match content {
+                Some(c) if c.contains("workspace = true") => {
```
sepointon, line 525: This does not look reliable.

crate2nix/src/resolve.rs
```diff
@@ -702,6 +724,314 @@ impl Display for NixSource {
     }
 }
 
+/// Reconstruct a standalone Cargo.toml from a `cargo_metadata::Package`.
+///
+/// When cargo resolves workspace inheritance, the `Package` struct contains
+/// all resolved values. This function serializes those back into a valid
+/// standalone Cargo.toml that doesn't reference any workspace.
+fn reconstruct_cargo_toml(package: &Package) -> String {
```
sepointon, line 732: I don't think this function is wrong, but I'm surprised by it having to be defined here. e.g., crates.io does reconstruction of Cargo.toml and must be using a similar function somewhere. Defining it in crate2nix seems like it'll become an implementation drift hazard. But maybe there's no existing library code that can be used for this and this really is the best option...

crate2nix/src/resolve.rs
```diff
@@ -702,6 +724,314 @@ impl Display for NixSource {
...
+/// containing a `Cargo.toml` with a `[workspace]` section. Returns the
+/// relative path from that workspace root to the crate's directory, or
+/// `None` if the crate is at the workspace root.
+fn find_git_workspace_sub_dir(manifest_path: &Path) -> Option<String> {
+    let crate_dir = manifest_path.parent()?;
+    let mut current = crate_dir.parent()?;
+    loop {
+        let candidate = current.join("Cargo.toml");
+        if let Ok(content) = std::fs::read_to_string(&candidate) {
+            if content.contains("[workspace]") {
```
sepointon, line 965: Also doesn't look reliable.

crate2nix/src/resolve.rs
```diff
@@ -702,6 +724,314 @@ impl Display for NixSource {
...
+    toml::to_string(&Value::Table(doc)).expect("failed to serialize reconstructed Cargo.toml")
+}
+
+/// Find the subdirectory of a crate within its git workspace.
+///
+/// Walks up from the crate's manifest path looking for a parent directory
+/// containing a `Cargo.toml` with a `[workspace]` section. Returns the
+/// relative path from that workspace root to the crate's directory, or
+/// `None` if the crate is at the workspace root.
+fn find_git_workspace_sub_dir(manifest_path: &Path) -> Option<String> {
```
sepointon, line 959: Other code with similar functions has had security vulnerabilities due to being tricked with symlinks/bind mounts into going into places they shouldn't. Is that a worry here as well?

crate2nix/src/resolve.rs
```diff
@@ -990,3 +1320,369 @@ pub struct ResolvedDependency {
     /// Extra-enabled features.
     pub features: Vec<String>,
 }
+
+#[cfg(test)]
+mod reconstruct_cargo_toml_tests {
```
sepointon, line 1325: Wishlist: if crate2nix really does need the renderer, then some fuzzing/property tests would be nice to have; at a minimum, testing round-tripping would be a very nice safeguard.

crate2nix/templates/nix/crate2nix/default.nix
```diff
@@ -419,15 +419,27 @@ rec {
...
+                        assert (lib.assertMsg (crateConfig ? sha256) "Missing sha256 for ${name}");
+                        crateConfig.sha256;
+                    });
+                in
+                if crateConfig ? resolvedCargoToml then
+                  let subDir = crateConfig.resolvedCargoTomlSubDir or ""; in
+                  pkgs.runCommand "${crateConfig.crateName}-${crateConfig.version}-src" { } ''
+                    cp -a ${rawSrc} $out
+                    chmod -R u+w $out
+                    cp ${builtins.toFile "Cargo.toml" crateConfig.resolvedCargoToml} $out/${subDir}/Cargo.toml
```
sepointon, line 439: Whose control is `subDir` under? May need quoting for shell meta-characters. If it's the crate author's, then there are potentially directory traversal attacks.

tools.nix
```diff
@@ -432,13 +494,42 @@ rec {
...
+                  shopt -u dotglob
+                  cp ${patchedCargoTomlFile} $out/Cargo.toml
+                  echo '{"package":null,"files":{}}' > $out/.cargo-checksum.json
+                ''
+              else
+                ''
+                  mkdir -p $out
+                  cp -apR ${src}/${pathToExtract}/* $out
+                  echo '{"package":null,"files":{}}' > $out/.cargo-checksum.json
+                '');
```
sepointon, line 533: Not actually sure if this would be an improvement, but maybe copying everything over and then replacing the Cargo.toml if needed afterwards would work better.

tools.nix
```diff
@@ -432,13 +494,42 @@ rec {
...
+              # Check if this crate uses workspace inheritance
+              crateCargoTomlParsed =
+                builtins.fromTOML
+                  (builtins.readFile "${src}/${pathToExtract}/Cargo.toml");
+              usesWorkspaceInheritance =
+                isWorkspace && hasWorkspaceTrue crateCargoTomlParsed;
+
+              # Resolve workspace inheritance by inlining dependency specs
+              resolvedCargoToml =
+                resolveWorkspaceInheritance crateCargoTomlParsed rootCargo;
```
sepointon, line 507: A second implementation of workspace inheritance in Nix makes me nervous for divergences and related bugs. Is there some reason this code can't read directly from the resolved Cargo.toml we're reading out of cargo-metadata and then researialising?

PR #445, opened 2026-03-11
experimental: add support for running clippy

### LINKED ISSUE #195: How to run clippy in a check
Making a derivation that runs `cargo clippy` fails due to trying to download the dependencies.

Looking at the clippy page it seems it can be run as `rustc` using `clippy-driver`. Is there a way with the `crate2nix` logic to follow this through or does something need to be added to `buildRustCrate` in nixpkgs?

If there is another nice way to run clippy in a derivation (looking at it for a flake check) then that would also be great.

jrobsonchase:
Closes #195 

# `runClippy` support

Add support for running clippy against a crate.

The user-facing API is similar to how `runTests` currently works. If `runClippy = true` in the call to `buildRustCrateWithFeatures`, an additional derivation is built that runs clippy against the root crate and is included in the "checks" (formerly "linked") derivation for the crate.

## Overall approach

As suggested in #195, clippy is invoked by replacing `rustc` with (a script that runs) `clippy-driver`. This also has to do a bit of argument massaging to ensure that clippy is run without `--cap-lints allow` and to drop back to rustc if `--version` is provided, since apparently some build scripts like to call `rustc --version` and get mad if clippy responds.

The clever bit of this is that rustc is _only_ replaced with clippy-driver in the root crate derivation, so all of its dependencies should be built with the originally-provided rustc and thus shouldn't need to be re-built for the clippy check. This is similar to how cargo-clippy works with the `--no-deps` argument.

## Design notes

* clippy _must_ be included in the toolchain provided as the `rustc` input to `buildRustCrate`.
  * Might be worth adding an additional clippy argument to `buildRustCrateWithFeatures`? But then it feels "different" from how `rustc` and `cargo` are provided in the `buildRustCrateForPkgs`, which doesn't feel great. I think that clippy is also tightly coupled to rustc, so if there's a version mismatch, users could have a bad time.
* Clippy is always run against the version of the crate derivation where `runTests = true`.
  * ~~I'm not entirely sure if this also implies that clippy is run against the crate's binary/library target as well, or if passing `--test` makes clippy _only_ run against the tests. Probably need to test some assumptions there.~~ Assumptions tested - clippy does run against all targets, lib, bin, and integration test with the current setup.
  * Maybe we want another `runClippyTests` flag so that clippy can be run both against the main crate and the test crate? This is only useful if users specifically *don't* want clippy to run against tests, which seems puntable until someone specifically needs it for some reason.

## Additional Changes

- Refactored `crateWithTests` into a function that can support any number of arbitrary checks, not just the running of tests. 
  - As a result, `runCrateTests` is now a standalone function that just constructs the "run the tests" derivation.
  - It also does the "Are there even any checks to include?" check, which was pushed down from `buildRustCrateWithFeatures`.
- Pushed the `buildTests = true` attribute down into `builtRustCratesWithFeatures` where it was previously overridden in `crateWithTests`.  I believe this is semantically equivalent.

## TODO

- [x] Documentation
- [x] Tests


Mic92:
@lovesegfault is this something you can apply in your internal test?

emanueljg:
@jrobsonchase I'm trying this out with a `Cargo.json` as generated by `crate2nix generate --format json` but it doesn't seem to work. I'm assuming `lib/build-from-json.nix` hasn't been updated in your PR?

jrobsonchase:
Nope, it has not. Are you saying that you can't enable clippy with `lib/build-from-json.nix` or that a regular build using a `Cargo.json` is broken with this change?

emanueljg:
> Nope, it has not. Are you saying that you can't enable clippy with `lib/build-from-json.nix` or that a regular build using a `Cargo.json` is broken with this change?

The former, there's no way to enable clippy.

```nix
            let
              cargoNix = import "${crate2nix}/lib/build-from-json.nix" {
                inherit pkgs;
                src = ./.;
                resolvedJson = ./Cargo.json;
                buildRustCrateForPkgs =
                  pkgs:
                  pkgs.buildRustCrate.override {
                    rustc = rust-bin;
                  };
              };
              # cargoNix = pkgs.callPackage ./Cargo.nix {
              #   buildRustCrateForPkgs =
              #     pkgs:
              #     pkgs.buildRustCrate.override {
              #       rustc = rust-bin;
              #     };
              # };
            in
            cargoNix.rootCrate.build.override {
              runClippy = true;
              features = [
                # ...
              ];

```
```
error:
       … while calling a functor (an attribute set with a '__functor' attribute)
         at /home/ejg/sia/flake.nix:154:13:
          153|             in
          154|             cargoNix.rootCrate.build.override {
             |             ^
          155|               runTests = true;

       (...)
      
       error: function 'anonymous lambda' called with unexpected argument 'runClippy'
       at «github:NixOS/nixpkgs/6dedf69f94d03cbe7bdde106f2d4c23ae2a853bf?narHash=sha256-gOTcX/9MZVMUE0Xvb4IEcv%2B0TQJkZFNEnL757ljU360%3D»/pkgs/build-support/rust/build-rust-crate/default.nix:160:5:
          159|     # Default: pkgs.rustc
          160|     {
             |     ^
          161|       rust ? rustc,
```

Now that I actually try to add `runTests = true` as well, I'm actually seeing that that is broken too in the same way. I'm not sure how that can be possible, but it'd mean that this error is out-of-scope for this PR. Somehow there are duplicate implementations of the crate build?

PR #484, opened 2026-06-01
test: set dontInstall

jmbaur:
Using `buildPhase` means that `stdenvNoCC.mkDerivation` still runs the other default phases, which in the default mode of operation can fail if a Makefile happens to exist in the directory the tests are ran from. Since we are only interested in running the tests with rustc, we can set `dontInstall` to true to prevent potential Makefile usage.


jmbaur:
cc @P-E-Meunier I saw you had made the last edits around this file

PR #486, opened 2026-06-16
fix: don't synthesize a phantom `default` feature for crates that declare none

### LINKED ISSUE #485: Crates without a `default` feature are built twice — `usesDefaultFeatures` injects a phantom `--cfg feature="default"`
# Crates without a `default` feature are built twice — `usesDefaultFeatures` injects a phantom `--cfg feature="default"`

## Summary

When a crate **defines no `default` feature** (no `[features]` table, or a `[features]` table without a `default` key) and is depended on from two edges — one with default features on, one with `default-features = false` — crate2nix generates **two distinct derivations** for it. They are byte-for-byte identical except that one is compiled with `--cfg feature="default"` and the other isn't.

In Cargo this is a no-op: `default-features = true` on a crate that declares no `default` feature enables nothing, so Cargo builds it once. crate2nix instead unconditionally adds `"default"` to the feature set for any default-features-on edge, regardless of whether the target crate actually declares that feature. The phantom feature changes the crate's `-C metadata` hash, forking its derivation — and because the fork is an *input* to everything downstream, it ripples up and duplicates the build of every crate in the dependency chain.

## Minimal reproduction

A workspace with a feature-less `leaf` crate, consumed once with default features on and once off:

```
c2n-repro/
├── Cargo.toml          # [workspace] members = ["leaf", "bin_a", "bin_b"]
├── leaf/Cargo.toml     # no [features] table at all
├── leaf/src/lib.rs
├── bin_a/Cargo.toml    # leaf = { path = "../leaf" }                        (default-features ON)
├── bin_a/src/main.rs
├── bin_b/Cargo.toml    # leaf = { path = "../leaf", default-features = false }  (OFF)
└── bin_b/src/main.rs
```

`leaf/Cargo.toml`:
```toml
[package]
name = "leaf"
version = "0.1.0"
edition = "2021"

# deliberately no [features] table — leaf defines no `default` feature
[dependencies]
```

Then:
```bash
cargo generate-lockfile
crate2nix generate
nix eval --impure --raw --expr '(import ./Cargo.nix { pkgs = import <nixpkgs> {}; }).workspaceMembers."bin_a".build.drvPath'
nix eval --impure --raw --expr '(import ./Cargo.nix { pkgs = import <nixpkgs> {}; }).workspaceMembers."bin_b".build.drvPath'
```

## Observed behavior

The `leaf` derivation differs between the two binaries:

```
bin_a (default-features ON ) leaf:  ldis7dx…-rust_leaf-0.1.0.drv   crateFeatures = "default"   -C metadata=dd84dd5913
bin_b (default-features OFF) leaf:  m1nnfm2…-rust_leaf-0.1.0.drv   crateFeatures = ""          -C metadata=1d8c657419
```

Two derivations for one feature-less crate; the only difference is the phantom `--cfg feature="default"` and the resulting metadata hash.

## Expected behavior

`leaf` should produce a single derivation (Cargo enables no feature for `default-features = true` on a crate with no `default` feature), so `crateFeatures` should be `""` on both edges.

## Root cause

`crate2nix/templates/nix/crate2nix/default.nix`, in `dependencyFeatures`:

```nix
dependencyFeatures =
  features: dependency:
    ...
    let
      defaultOrNil = if dependency.usesDefaultFeatures or true then [ "default" ] else [ ];
      ...
```

`defaultOrNil` adds `"default"` based solely on the **edge's** `usesDefaultFeatures`, without checking whether the **target crate** actually declares a `default` feature. For a crate whose generated record has no `features` attr (or one lacking a `default` key), this injects a feature that doesn't exist.

## Suggested fix

Only inject `"default"` when the target crate actually declares it — i.e. gate `defaultOrNil` on the target crate config containing a `default` feature (the resolved crate configs are already available to the caller, keyed by `dependency.packageId`):

```nix
defaultOrNil =
  if (dependency.usesDefaultFeatures or true)
     && (crateConfigs.${dependency.packageId}.features or {}) ? "default"
  then [ "default" ]
  else [ ];
```

This mirrors Cargo's semantics exactly (`default-features = true` is a no-op when no `default` feature exists).

## Impact

In a real workspace this is not cosmetic. We hit it on `dotenvy`, `equivalent` (a crate with *no* `[features]` table), `icu_locale_core`, `jsonc-parser`, `reqwest-middleware`, and `secrecy`. Each fork rippled up its dependents — e.g. `equivalent`/`icu_locale_core` forked the entire `sqlx` stack (`sqlx-core`, `sqlx`, `sqlx-postgres`, `sqlx-macros`, `sqlx-macros-core`), which in turn compiled large first-party crates 2–4× across the binary set. Collapsing the phantom-`default` forks removed dozens of redundant compiles.

## Workaround

Pin the affected crate to the feature set Cargo actually resolves, via a crate override:

```nix
# in the crateOverrides passed to the generated Cargo.nix
equivalent = _: { features = [ ]; };
icu_locale_core = _: { features = [ "alloc" "serde" "zerovec" ]; };
```

## Environment

- crate2nix 0.15.0 (latest release, 2026-01-28)
- `buildRustCrate` from nixpkgs
- Reproduced on `x86_64-linux`

---

## Update: the empty `default = []` edge case

The suggested fix above (and the first revision of #486) strips the phantom `"default"` whenever the target crate has no `default` key in its generated `features` map. That is correct for a crate that declares **no** `default` feature, but it over-reaches for a crate that declares an **empty** `default = []`: crate2nix omits an empty default from `crateConfig.features`, so the `? "default"` test can't tell the two apart, and the strip wrongly hits the empty-default crates too.

An empty `default = []` enables no other features, but the crate's **source can still gate on the cfg**. The canonical example is [`document-features`](https://docs.rs/document-features):

```toml
[features]
default = []
```
```rust
#[cfg(not(feature = "default"))]
compile_error!("The feature `default` must be enabled to ensure \
    forward compatibility with future version of this crate");
```

Under Cargo, `default` stays on (nobody disables it), so the `compile_error!` never fires. Stripping `"default"` in crate2nix turns the cfg off and the crate fails to compile — which breaks every build that pulls it in (e.g. `crossterm → document-features`).

The fix in #486 therefore distinguishes **three** cases instead of two, using `resolvedDefaultFeatures` (which records an empty default that `features` omits):

1. non-empty `default = [ ... ]` → leave the requested features alone;
2. empty `default = []` (present in `resolvedDefaultFeatures`, absent from `features`) → force `"default"` **on** everywhere — keeps the cfg satisfied and the derivation unique (`resolvedDefaultFeatures` is Cargo's single global resolution, so it's consistent across every closure);
3. no `default` at all → strip the phantom (the original fix).

Note this also means the per-crate `features = [ ... ]` override workaround is only safe for case-3 crates (`equivalent`, `icu_locale_core`, …). Applying it to an empty-default crate that gates on the cfg (like `document-features`) would reintroduce the miscompile.

varunshahdev:
Fixes #485.

## Problem

`mergePackageFeatures` resolves every crate's features via:

```nix
expandedFeatures = expandFeatures (crateConfig.features or { }) features;
```

where `features` is the requested set — a dependency's resolved features (which include `"default"` for any `usesDefaultFeatures` edge, via `dependencyFeatures`) or the root's `rootFeatures` (default `[ "default" ]`). It keeps `"default"` even when the target crate **declares no `default` feature**.

Cargo treats requesting `default` on such a crate as a no-op (it enables nothing). crate2nix instead stamps a phantom `--cfg feature="default"`, which changes the crate's `-C metadata` hash. So a crate reached both **with** and **without** `"default"` — a default-on dep edge vs a `default-features = false` edge, or a root build vs a dependency — forks into **two derivations**, and the fork propagates to everything downstream. Crates with no `[features]` table at all (e.g. `equivalent`) are the clearest trigger; see #485 for a minimal repro (a feature-less `leaf` crate compiled twice).

## Fix

Reconcile the requested feature set at this single chokepoint with what Cargo actually resolves. There are **three** cases, not two, because crate2nix omits an *empty* `default = []` from `crateConfig.features` — so the simple `? "default"` test cannot tell "declares no default" apart from "declares an empty default":

```nix
requestedFeatures =
  if (crateConfig.features or { }) ? "default" then
    # 1. Non-empty `default = [ ... ]`: Cargo resolves it normally; pass through.
    features
  else if builtins.elem "default" (crateConfig.resolvedDefaultFeatures or [ ]) then
    # 2. Empty `default = []`: omitted from `features` but recorded in
    #    `resolvedDefaultFeatures`. Enables nothing, but the crate's source may
    #    gate on `cfg(feature = "default")`, so keep it — force it on everywhere.
    features ++ [ "default" ]
  else
    # 3. No `default` at all: strip the phantom, collapsing the fork.
    lib.filter (f: f != "default") features;
expandedFeatures = expandFeatures (crateConfig.features or { }) requestedFeatures;
```

Case 2 is the important correction. An earlier revision of this PR stripped `"default"` whenever it was absent from `crateConfig.features`, on the assumption that an empty `default = []` is always a pure no-op. It isn't: [`document-features`](https://docs.rs/document-features) declares `default = []` yet carries `#[cfg(not(feature = "default"))] compile_error!(...)`, so stripping `"default"` makes it fail to compile under crate2nix while it builds fine under Cargo (which leaves default features on). Forcing the empty default **on** rather than off both satisfies the cfg and keeps the derivation unique — `resolvedDefaultFeatures` is Cargo's single global resolution, so the feature set is identical in every closure that reaches the crate, and consistently-on dedups exactly as consistently-off would.

## Tests

Updates the `packageFeatures` tests whose expectations encoded the old behavior — a crate that declares no `default` feature, requested with `"default"` (via a dep edge **or** as a root feature), now resolves without it. Adds a regression test for case 2: a crate with an empty (resolved) default reached only through a `default-features = false` edge keeps `"default"`. `lib.runTests` over `templates/nix/crate2nix/tests` returns `[]` (all pass).

Verified end-to-end against the #485 repro (the two `leaf` derivations collapse to one), against `document-features` (it compiles again), and in a large real workspace (the whole sqlx stack and domain layer de-duplicate, standalone binaries stop forking from their in-dependency copies, and a drv-count shows zero same-name+version forks).

### DIFF
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -584,17 +584,40 @@ rec {
       assert (builtins.isBool runTests);
       let
         crateConfig = crateConfigs."${packageId}" or (builtins.throw "Package not found: ${packageId}");
-        # Cargo treats `default-features = true` on a crate that declares no
-        # `default` feature as a no-op (it enables nothing). crate2nix otherwise
-        # keeps the synthesized "default" — from a default-features-on dependency
-        # edge or from the root's default `rootFeatures = [ "default" ]` — and
-        # stamps a phantom `--cfg feature="default"`, forking the crate's
-        # derivation (different `-C metadata`) from any path that reaches it
-        # without "default". Strip it here, the single point every crate's
-        # feature set flows through, so both cases are covered at once.
+        # crate2nix synthesizes "default" for a crate whenever it is reached with
+        # default features on — from a default-features-on dependency edge or from
+        # the root's default `rootFeatures = [ "default" ]` — and stamps a
+        # `--cfg feature="default"` that changes the crate's `-C metadata` hash.
+        # When the crate is *also* reached without "default", it forks into two
+        # derivations and the fork propagates to everything downstream. We
+        # reconcile the requested feature set with what Cargo actually resolves,
+        # which depends on whether — and how — the crate declares `default`:
+        #
+        #   1. Declares a non-empty `default = [ ... ]`
+        #      (`crateConfig.features ? "default"`): Cargo resolves it normally
+        #      and crate2nix already matches — pass `features` through untouched.
+        #
+        #   2. Declares an *empty* `default = []`: crate2nix omits an empty
+        #      default from `crateConfig.features` (so case 1's test misses it),
+        #      but records it in `crateConfig.resolvedDefaultFeatures`. "default"
+        #      enables no extra features, yet the crate's source may still gate on
+        #      `cfg(feature = "default")` (e.g. document-features carries
+        #      `#[cfg(not(feature = "default"))] compile_error!(...)`), so we must
+        #      NOT strip it. `resolvedDefaultFeatures` is Cargo's single global
+        #      resolution, so we force "default" on in every closure that reaches
+        #      the crate: the cfg stays satisfied and the derivation stays unique
+        #      (consistently-on dedups exactly as consistently-off would).
+        #
+        #   3. Declares no `default` at all (absent from both `features` and
+        #      `resolvedDefaultFeatures`): Cargo never sets `cfg(feature =
+        #      "default")`, so the synthesized "default" is a phantom — strip it,
+        #      collapsing the fork. `default-features = true` on such a crate is a
+        #      Cargo no-op.
         requestedFeatures =
           if (crateConfig.features or { }) ? "default" then
             features
+          else if builtins.elem "default" (crateConfig.resolvedDefaultFeatures or [ ]) then
+            features ++ [ "default" ]
           else
             lib.filter (f: f != "default") features;
         expandedFeatures = expandFeatures (crateConfig.features or { }) requestedFeatures;
--- a/crate2nix/templates/nix/crate2nix/default.nix
+++ b/crate2nix/templates/nix/crate2nix/default.nix
@@ -584,7 +584,20 @@ rec {
       assert (builtins.isBool runTests);
       let
         crateConfig = crateConfigs."${packageId}" or (builtins.throw "Package not found: ${packageId}");
-        expandedFeatures = expandFeatures (crateConfig.features or { }) features;
+        # Cargo treats `default-features = true` on a crate that declares no
+        # `default` feature as a no-op (it enables nothing). crate2nix otherwise
+        # keeps the synthesized "default" — from a default-features-on dependency
+        # edge or from the root's default `rootFeatures = [ "default" ]` — and
+        # stamps a phantom `--cfg feature="default"`, forking the crate's
+        # derivation (different `-C metadata`) from any path that reaches it
+        # without "default". Strip it here, the single point every crate's
+        # feature set flows through, so both cases are covered at once.
+        requestedFeatures =
+          if (crateConfig.features or { }) ? "default" then
+            features
+          else
+            lib.filter (f: f != "default") features;
+        expandedFeatures = expandFeatures (crateConfig.features or { }) requestedFeatures;
         enabledFeatures = enableFeatures (crateConfig.dependencies or [ ]) expandedFeatures;
         depWithResolvedFeatures =
           dependency:
--- a/crate2nix/templates/nix/crate2nix/tests/packageFeatures.nix
+++ b/crate2nix/templates/nix/crate2nix/tests/packageFeatures.nix
@@ -53,6 +53,19 @@ let
       features = { };
     };
 
+    # A crate that declares no `default` feature, depended on with default
+    # features left on (the implicit default for the edge). Cargo enables
+    # nothing in this case, so neither should crate2nix.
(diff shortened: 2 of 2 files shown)


varunshahdev:
Pushed a follow-up commit (`ac2f0d9`) correcting an edge case: an **empty** `default = []` is omitted from `crateConfig.features`, so the original two-way strip also caught crates that declare an empty default. That's wrong when the crate's source gates on `cfg(feature = "default")` — `document-features` declares `default = []` *and* carries `#[cfg(not(feature = "default"))] compile_error!(...)`, so stripping made it fail to compile (it builds fine under Cargo, which leaves default features on).

The fix now uses `resolvedDefaultFeatures` to split three cases: pass a non-empty declared default through, force an empty (resolved) default **on** everywhere (keeps the cfg satisfied *and* the derivation unique), and strip the phantom only when no default is declared at all. Updated the PR description and added a `packageFeatures` regression test. Details in #485.

domenkozar:
Can you check why CI fails?

